Issue #94, opened 2026-09-17
[dev] Day-relative chips go stale at midnight: a pick after midnight acts on a different day (drift shift, rescue shift/split/roll, day-load trim)


Derpimort:
## What happens
A chip's reply is plain chat that is parsed **when it's picked**, and chips stay live until a newer user message (`choicesActive` has no day check). Several chips name their day with **relative** words, so a pick after midnight speaks about a different day than the one offered:

| Chip | Where | Reply shape |
|---|---|---|
| drift **shift** (#12/#63) | `store.ts` `offerDriftChoices` | `move the release review at 14:00 to tomorrow at 9:00` |
| rescue **shift** | `domain/rescue.ts:137` | `move the <block> to <today/tomorrow/weekday> at <time>` |
| rescue **split** | `domain/rescue.ts:152` | `splitReply(…, day)` |
| rescue **roll** | `domain/rescue.ts:162` | `move the <block> to <tomorrow/weekday>` |
| day-load **trim** (#301) | `domain/insights.ts:873` | `move the <block> to <word>` |

Example (from the parse, not yet probed through the store): offered Tuesday, "move the release review at 14:00 to tomorrow at 9:00" picked at Wednesday 00:05 parses "tomorrow" as **Thursday**. The offer only checked Wednesday 9:00 for a clean slot.

The **drift drop** chip had the same staleness plus a wrong-block removal. #89 fixes it with a pick-time re-check (`dropReplySinglesOut` in `pickChoice`), scoped to `drop-<id>` chips only. The rest are listed here, as the manager ruled on #89 (the shift chip's id `shift` is shared by drift and rescue, so a drift-only re-check can't tell them apart at pick time).

## Acceptance
- A day-relative chip picked on a later calendar day than its offer (`msg.ts`) either:
  - re-resolves to exactly what it offered (same block, same landing day and time), or
  - is spent with a kind line naming what it was for, with nothing moved.
- Same-day picks are unchanged.
- One rule for every chip kind above, not one per offer. Candidates:
  - a `dayKey(msg.ts) !== today` check, plus a re-derivation of the reply against the offer's day;
  - or storing the offer's resolved target on the choice.
- Pins through the real store for each chip kind: offer on Tue, pick at Wed 00:05 → nothing moves.

## Laws
- Offer-then-confirm: a pick acts only on what the owner saw.
- Positive voice: the kind line names what the choice was for.

[...]

Issue #75, opened 2026-09-17
[dev] Batch changes with a confirm before a wide blast: the batch op from #16's calendar command surface


Derpimort:
## What & why

Split from #16 (calendar command surface). v0.6 shipped resize, duplicate and relative move (Derpimort/mew-archive#353). **Batch** is the third remaining op, and the one with a blast radius.

The owner wants to say *"push everything after 3pm back an hour"* or *"move all of today's admin to tomorrow"*. Today the model has to call `move_relative` / `move_task` once per block, with no consent step for a wide change and one undo snapshot per call.

## Architecture placement
- `adapters/model/tools.ts` → a `batch_blocks` tool: a **selector** (`day`, `afterMin` / `beforeMin`, `tag`, `titleQuery`) plus **one op** (`shift` by minutes, `move_to_day`, `set_tag`) and a `confirmed` flag.
- `state/store.ts` → `execBatch`:
  1. Resolve the selector with the same resolver `list_blocks` uses (`execListBlocks`), so what's listed is exactly what's touched.
  2. If more than a small threshold (e.g. 3 blocks), or any block on another day, post **`offer_choices`** naming the count and the span ("move 5 blocks after 15:00 back an hour?" · *do it* / *not now*). **Nothing changes before the pick.**
  3. On the pick, apply every change through `setBlocks` as **one** undo snapshot.
- `domain/week.ts` → a pure `planBatch(blocks, selector, op)` → the moves plus the blocks it must skip, with the reason for each.
- A keyless route for the two commonest phrasings ("push everything after <time> back <n>", "move all <tag> to tomorrow").

## Product laws
- Offer-then-confirm for wide changes (consent for blast radius); a single-target op keeps acting directly.
- Tools-only; one mutation path; one undo for the whole batch.
- **External events are never moved** by a batch, and fixed-time blocks are scheduled around, never over. The confirm names what stays put.
- Positive voice.

## Acceptance criteria
- [ ] "push everything after 3pm back an hour" → one confirm chip naming the count; nothing moves until the pick; the pick moves exactly the listed blocks as one undo; pinned (keyless and keyed)
- [ ] `[calendar]` and fixed blocks in the selection are skipped and named in the confirm; pinned
- [ ] A 1–2-block selection on the same day acts directly, like today's single ops; pinned
- [ ] A collision after the shift uses the existing clash/drift wording (no second vocabulary)
- [ ] tsc · coverage · lint · format · build green; CHANGELOG line

## Out of scope
[...]

PR #141, opened 2026-09-17
fix(clash): the clash note never calls a protected rest "flexible" — it names it the way #122 does

Derpimort:
## What
Task `tsk_01m2r6t6epkkex1zgq2qapx01e`, from my #137 review 🟢. Off RC `4b3bb6d`, with #137's rest wording in.

A change that landed on a **protected rest** through `clashNote` called that rest **flexible**. Two examples:
- `push all work after 5pm back 30 min` → `Moved 1 block 30 min later today — Deck 17:00→17:30. — note: it overlaps Evening walk 18:00–18:45 (flexible)`
- A keyed model was told `(flexible — offer to drift it, don't move it unasked)`.

A protected rest isn't flexible. It's sacred time that protect-rest owns, and a placement never moves it (#324, #122).

## How
`clashNote` (`state/store.ts`) now splits a clash three ways:

| Block | Wording |
|---|---|
| **Fixed-time** | `(fixed — it can't move)`, unchanged |
| **Truly flexible** | `(flexible)`; for the keyed model also `DRIFT_OFFER_NOTE`, which #119's `ownerView` keeps off the floor. Unchanged |
| **Protected rest** (tag rest, protected, after the fixed-time check, the same order `driftCollisions` uses) | #122's words: ` — it runs over your evening walk 18:00–18:45`, with **no offer to drift** |

- **One helper, `restRunsOver`,** now builds that phrase for both `clashNote` and #137's `driftReply`, so the two paths can't drift apart. `isProtectedRest` is the one rule.
- **Every `clashNote` path gets it:** batch receipts, edits, duplicates, and the plan/move clash for non-work blocks.

## Tests
New `state/__tests__/clash-note-rest.test.ts` (5):
1. **The batch repro:** `Moved 1 block 30 min later today — Deck 17:00→17:30. — it runs over your evening walk 18:00–18:45`, and the walk stays at 18:00–18:45.
2. **A truly flexible block keeps its note:** `… — note: it overlaps Gym 18:00–19:00 (flexible)`.
3. **Both at once:** `… — note: it overlaps Gym 18:00–18:30 (flexible) — it runs over your evening walk 18:30–19:15`.
4. **Keyed batch results:** onto the walk, `… — it runs over your evening walk 18:00–18:45` with no `offer to drift`; onto Gym, `(flexible — offer to drift it, don't move it unasked)` is kept for the model.
[...]

PR #140, opened 2026-09-17
fix(remove): a typed answer to a live remove ask does what its chip does — "both", "all of them", "the thursday one" (#131)

Derpimort:
## What
Fixes #131 (task `tsk_01m2r5nqpvc3zpcqw28gxat93r`, coderpb's queue item, taken by mutual agreement). Off RC `4b3bb6d`, with #130 and #123 in.

The remove ask tells the owner to **say** `"both"` / `"all of them"`, but on the keyless floor typing it did nothing useful, and only the chip worked:
- `all of them` got the generic `I can place that for you — try …` line;
- `both` and `all` became **inbox thoughts** called `both` / `all`;
- `remove all of them` became a thought called **`remove`**.

## How
- **`domain/choices.ts`, `typedRemoveAnswer(chat, text)` (pure).** It applies while the **newest live** chips message is a **remove ask**, meaning every option's reply removes something. A whole-message reply that points at **exactly one** chip resolves to it:
  - the chip's **label** (`thursday 12:00`, `the 12:30`, `both`);
  - the **all-chip's words**: `both`, `all`, `all of them`, `them all`, `all three`, and each of those after `remove`;
  - a **day**: `the thursday one`, `thursday's`, `the one on thursday`, `tomorrow`;
  - a **time**: `the 12:00`, `12:00`.

  An optional leading `remove` / `drop` / `delete` is allowed. Anything else, including a picked ask, an ask left behind by a newer message, a non-remove chip family, or words shared by two chips, gives `null`.
- **`state/store.ts`.** `speak()` resolves a typed answer first and hands it to **`pickChoice`**, the tap's own path. So #94's pick-time re-check, the drop-chip check and the chips' retirement all apply. The log shows the chip's reply (`remove all lunch`), the same as a tap. Every other message is ordinary, exactly as before.

**Calls I made inside the laws:**
- **A series-scope remove ask has no all-chip, so `all` never means "the whole series".** Its scope chips (`just this one`, `this & the ones after`, `the whole series`) all remove too. The all-words only resolve to a chip whose reply is `remove all …`. The exact label `the whole series`, typed, still picks it.
- **The log shows the chip's reply, not the typed word.** Going through the tap's path keeps a single path and a single set of checks, and the owner sees exactly how their words were read.
- **The copy keeps "say".** The words it names now work when typed, so "pick" isn't needed.
- **No live ask means nothing changes.** `both` on its own is still captured, as before; the brief asked that behaviour stay.

[...]

PR #142, opened 2026-09-17
feat(batch): retag a selection in place, and 'between X and Y' windows (#75 slice 2)

Derpimort:
Refs #75 (**slice 2**; the issue stays open for the series slice). Task `tsk_01m2r14b1c7fb62gb0tjf01wa8`. Head `df68e34` on RC `84e93df`. Commits:
- `4d11227`: the slice
- `348ab04`: two wording fixes from the live probe
- RC merges

## What the owner can say now
- **`tag all of tomorrow's calls as work`** · `retag all today's "gym" after 17:00 as health` · `mark all thursday's private as work`
- **`push everything between 2 and 5pm back 30 min`** · `move all today's work between 14:00 and 17:00 to tomorrow`
- Keyed models get the same through `batch_blocks` `op: 'set_tag'` + `toTag`.

```
you ❯ tag all of tomorrow's calls as work
mew ❯ tag 3 blocks as work tomorrow? Client call 9:00 · Vendor call 11:00 · Call the bank 13:00. Board call 15:00 (from your calendar) keeps its tag.
      [do it]  [not now]
you ❯ tag all tomorrow's "calls" as work — yes, all 3 · …                     (tap)
mew ❯ Tagged 3 blocks as work tomorrow — Client call 9:00 · Vendor call 11:00 · Call the bank 13:00. Board call 15:00 (from your calendar) keeps its tag.
```

## Slice 1's laws, unchanged for a retag
- **The exact list is offered first** for a wide selection (three or more); two act directly, like the single ops.
- **The yes** carries the count and the list token (now including the tag), and it re-offers if anything changed.
- **One undo.** A confirm picked after midnight re-checks (#94), and weekday-named confirms still act.
- **A retag changes only tags, never a time.**
- **Named and left alone:** calendar events, done blocks and repeating blocks keep their tags, and a block that already has the tag is named: `Walk 17:00 (already health) keeps its tag.`

## The change
- **`domain/batch.ts`:**
  - `BatchOp` gains `setTag`; `planBatch` plans it in place (no landing or off-day checks) with the new `already` skip;
  - `batchToken` includes the tag;
  - `selectBatch`: a **plural title word that matches nothing that day tries its singular** (`calls` → Client call). It never widens a selection that already matched.
- **`state/store.ts` `execBatch`:**
  - per-op wording (`tag N blocks as work tomorrow?` / `Tagged …` / `keeps its tag` / `nothing there can be tagged …`);
  - the retag re-ask (`tag all <day>'s <selection> [window] as <tag>`);
  - a retag moves nothing, so it shares no new time;
  - the retag's receipt card and working line say `tagging them`;
  - a two-edge window reads as one: `starting between 14:00 and 17:00`.
- **`domain/parse.ts`:**
  - `tag|retag|mark all [of] [day's] <selection> [window] as <tag>`, sharing `readSelection` with the move form (batch words still required, so `tag all hands as work` stays out);
  - `between X and Y` normalizes to the two-edge window. A bare first hour takes the second's half of the day when it comes earlier on the clock (`between 2 and 5pm` = 14:00–17:00) and the other half when it can't (`between 11 and 1pm` = 11:00–13:00).
- **`adapters/model`:** `batch_blocks` `set_tag` + `toTag`; the rules floor, `sanitizeIntent` and the executor type carry it; `toolCard` gains `retag`.
- **`domain/chipEffect.ts`:** the batch effect includes `toTag`.

## Calls I made
- **An own fixed-time block takes a tag.** The task's law says fixed blocks are never touched, and its own example (`tomorrow's calls`) is exactly the blocks `isFixedTime` reads as fixed by title. That law protects **time**: a fixed block is scheduled around, never moved or covered. A retag moves nothing and covers nothing, so the owner's own "Client call" takes the tag, while a move or shift still skips it. **Calendar events keep their tags** (someone else's event).
- **The plural fallback** only fires when the literal words match nothing that day, and the exact list is offered before any wide change.
- **Series-aware selection (part c)** ships as the next slice. The this / following / series scope chips need their own scope-ask and midnight re-check design; here, repeating blocks are named and left alone exactly as in slice 1.

## Proof
**Pure** (`domain/__tests__/batch-slice2.test.ts`, 17):
- `planBatch` retag: retags in place; calendar, done, repeating and already each named; an own fixed call takes a tag while a shift still skips it; the token differs by tag;
- the plural fallback (literal first; never for short words or when something matched);
- the grammar: retag forms including the yes with its token, `between` in three shapes, and `tag all hands as work` stays out;
- a retag confirm picked after midnight: the weekday form still means the same, the tomorrow form doesn't.

**Real store** (`state/__tests__/batch-slice2-store.test.ts`, 7):
- the exact offer, the reply and the receipt;
- two blocks act directly;
- all-already says so;
- a keyed offer tapped on the keyless floor retags exactly its list, and one undo in a keyed turn restores every tag;
- the `tagging them` card;
- `between 2 and 5pm` selects exactly the blocks starting 14:00–16:59, with its re-ask;
- nothing in the window gives the one-window line.

Slice 1's suites are unchanged apart from `toTag: null` in the two chip-effect shapes.

**Mutants** (each applied, run over the five batch suites plus the store file; restored):
```
S1  a retag also moves the block               ✘ 1    S9  plural falls back even when the literal matched  ✘ 1
S2  no "already" skip                          ✘ 2    S10 no plural fallback                               ✘ 2
S3  a retag skips own fixed blocks             ✘ 7    S11 the chip effect ignores the tag                  ✘ 2
S4  the token ignores the tag                  ✘ 1    S12 the re-ask says "move" for a retag               ✘ 2
S5  no tag grammar                             ✘ 8    S13 the floor drops the retag op                     ✘ 4
S6  "between" not normalized                   ✘ 5    S14 the retag card says "batch"                      ✘ 1
S7  "between" inherits no half of the day      ✘ 4    S15 two edges read as two phrases                    ✘ 1
S8  "between" swallows the trailing space      ✘ 1
```
**S8 was a real bug** caught while testing. The clock's optional am/pm let `17:00 ` swallow the following space, so `between 14:00 and 17:00 to tomorrow` glued into `17:00to` and parsed as a single-block move.

### Live, in the built app (keyless, `?d=2026-09-16`)
```
you ❯ block 30 min for probe call alpha tomorrow at 9     (and beta at 11, gamma at 13)
you ❯ tag all of tomorrow's probe calls as health
mew ❯ tag 3 blocks as health tomorrow? probe call alpha 9:00 · probe call beta 11:00 · probe call gamma 13:00.
you ❯ … as health — yes, all 3 · 1h4ym2l                                   (tap)
mew ❯ Tagged 3 blocks as health tomorrow — probe call alpha 9:00 · probe call beta 11:00 · probe call gamma 13:00.

you ❯ push everything between 2 and 5pm tomorrow back 30 min
mew ❯ nothing matches everything starting after 14:00 starting before 17:00 tomorrow, so everything stays as it is.
```
- **The fallback:** `probe calls` matched nothing literally, and its singular found the three.
- **Fixed after this probe (`348ab04`):** the retag's card read `✓ moving them`, and the two-edge line read `starting after 14:00 starting before 17:00`. Both are now `tagging them` / `starting between 14:00 and 17:00`, and both are pinned.
- **The probe itself:** it clicked the chip through the DOM, because Playwright's locator click on the chip timed out waiting for actionability. The chip was enabled, and the app is unchanged in that area.

Page errors: none. This was a throwaway Playwright probe against `pnpm preview`; it wasn't committed.

## Gates (from `app/`, head `df68e34` on RC `84e93df`, served on :5202)
```
cwd=/home/work/work/mew/.worktrees/rc1-75s2/app head=df68e34
tsc exit=0
 Test Files  172 passed | 2 skipped (174)
      Tests  2804 passed | 4 skipped (2808)
Lines        : 93.75% ( 7110/7584 )
coverage exit=0
lint exit=0
format exit=0
build exit=0
| ✅ | **main** (entry + its static non-vendor chunks) | | **392.8 KB** | **440.0 KB** |
| ✅ | **first load** (eager) | | **796.4 KB** | **1200.0 KB** |
bundle exit=0
shoot exit=0
overlap exit=0
```
Canon PNGs were restored after the shoot. CHANGELOG `[Unreleased]` → *Change several blocks at once, with a yes first* gains one bullet. `check-changelog.mjs` → `31 sections (31 unique) · 56 bullets (56 unique) — no repeats`.

**Laws:**
- **Offer-then-confirm** with the exact list;
- **calendar events, done and repeating blocks untouched;**
- **no time ever changes on a retag;**
- **one undo;**
- **positive voice;**
- **keyless parity.**


Derpimort:
**Peer review (coderpa) — APPROVE. 0 🟡, three 🟢 (one I'd take if you take any).**

Read @df68e34 in its own worktree (your merge of RC `84e93df`; RC is `131a56c` now, GitHub still MERGEABLE, and neither #141 nor #140 goes near these lines).

## What I checked

**The laws hold.** `external` is tested before anything else, so a calendar call is named and never retagged; repeating blocks and done blocks keep their tags; the offer comes first for 3+ blocks and for any day move, and a ≤2-block retag acting directly is exactly slice 1's shift precedent. `setTag` writes `{...b, tag}` and nothing else — no day, no minute — and `sharing` is emptied for a retag, so a receipt never invents a clash for blocks that didn't move.

**The fixed-time call is the right one, for the reason the comment gives.** Fixed-time is about *time*; a retag moves nothing, so an owner's "Client call" taking `work` doesn't touch what fixed-time protects. The move path still skips it, in the same `planBatch`.

**The token change closes the reuse hole:** `#tag` in `batchToken` means a yes to "as work" can't act on a plan that now says "as health".

**`readSelection` is a faithful extraction** — the `sel` → `named` rename aside, every branch (quoted, `blocks`, tag word, title words, `my`, the signal rule) is slice 1's, and the move path's `toDayOffset == null` check now sits before it with the same effect.

- `batch-slice2` + `batch-slice2-store` + `batch-blocks` + `chipEffect-batch` + `parse`: **132 passed**.
- Mutants (sources restored byte-identical after each) — **all six caught**:

| Mutant | Caught by |
|---|---|
| U1: the singular fallback widens even when the plural matched | "calls matches the literal first" |
| U2: a retag skips fixed-time blocks again | the retag pin + two skip pins |
| U3: a block that already has the tag is retagged anyway | the `already` pins, domain and store |
| U4: the list token forgets the tag | the token pin |
| U5: a bare first hour always takes the second's half | "between 11 and 1pm" |
| U6: the retag confirm re-asks in the move's words | both keyless and keyed offer pins |

**Probed the `between` grammar past its pins** (`parseCommand`, Tuesday 08:30):

| Said | Read as |
|---|---|
| `between 2:30 and 5pm` | 14:30 → 17:00 ✅ |
| `between noon and 3pm` | 12:00 → 15:00 ✅ |
| `between 10 and 2pm` | 10:00 → 14:00 ✅ |
| `between 9 and 11am` | 9:00 → 11:00 ✅ |
| `between 3 and 4` | the bare-hour ask ✅ (no guess) |
| `between 5pm and 8` | the bare-hour ask 🟢 |
| `between 11 and 2am` | after 23:00 **and** before 02:00 🟢 |
| `"the gap between 2 and 5pm"` as title words | title rewritten 🟢 |

## 🟢 1 — the one I'd take: the rewrite reaches inside quoted title words

```
move all "the gap between 2 and 5pm" blocks to friday
  → titleQuery: "the gap after 2pm and before 5pm"
```

`betweenAsEdges` runs over the whole line before any match, so a title that contains "between X and Y" is rewritten too: nothing is selected, and MEW echoes words the owner never said back at them ("nothing matches "the gap after 2pm and before 5pm""). Nothing is lost, which is why it's 🟢 — but the fix is small: skip quoted spans, e.g. rewrite only the parts outside `/"[^"]*"/` (split, map the odd indices through unchanged, rejoin).

## 🟢 2 — the inference is one-directional

`between 2 and 5pm` reads the bare first hour off the second edge, but `between 5pm and 8` has no rule for the bare *second* hour, so it falls to the ask — and the ask then suggests `say "after 8pm"` when 8 is the *before* edge (that wording is slice 1's `windowOf`, not yours; worth its own issue either way). Mirroring the inference — a bare `to` takes `from`'s half when the clock allows — would make the two edges symmetric.

## 🟢 3 — a window that can never match

`between 11 and 2am` → `afterMin 1380, beforeMin 120`, so the selection is empty by construction and MEW says "nothing matches everything starting between 23:00 and 2:00". Slice 1 can already reach that with "after 11pm and before 2am", so it isn't new — but the am/pm inference is a new way in. If `afterMin >= beforeMin`, asking ("11am to 2pm?") beats selecting nothing.

**One observation, no action:** the singular fallback catches `calls → call` but not `classes → classe`; conservative is right here, just worth knowing it's an -s rule, not a plural rule. And a seeded meal (`placedBy`) does take a retag — correct, I think: the owner named the selection and the offer lists the block by name.

Strong slice — the token, the skip reasons and the confirm round-trip all line up, and the test net caught everything I threw at it.

Issue #117, opened 2026-09-17
[dev] Keyless "tonight" is not read as the evening (lands 14:00–16:00), and "this evening" leaves "this" in the title


Derpimort:
## What happens
On the keyless floor, **"tonight"** isn't read as the evening, and **"this evening"** leaves "this" in the title.

**Repro (real store, keyless):** Tuesday 14:00, the afternoon and evening open.
- `block 2h for writing tonight` → `Done — today 14:00–16:00 is held for writing.`
  - The block sits in the **afternoon**. "tonight" is only stripped from the title (parse.ts:181), with no day and no window, and the scorer's evening damping then prefers the afternoon.
- `block 1h for reading this evening` → `Done — today 18:00–19:00 is held for reading this.`
  - The title is **"reading this"**. "evening" reads as a stated 18:00 start, and the "this" stays behind.

## Expected
- "tonight" / "this evening" places inside the evening (after the classic day end, inside the plannable hours), like the keyed `window: 'evening'` (#51 #53 #55).
- The day words don't leak into the title: `block 1h for reading this evening` → title "reading".

Found by the RC cross-feature journeys (`rc-journeys.test.ts`, test-only). It's pinned there with `it.fails`.

PR #89, opened 2026-09-17
feat(drift): the drop chip names its day, so a same-titled block elsewhere no longer hides it (#12 follow-up)

Derpimort:
The #12 follow-up. Task `tsk_01m2qd26kq8cdv7jq6c56sy0tc`. Base: RC `b1d8799`. This PR is one commit, `a3acee0`.

## What was left
#63's no-clean-drift ask offers **drop the flexible block**. Its exactness guard hid the chip whenever a same-titled block sat at the same time on another day, because the day-less reply (`remove the Groceries at 14:00`) could have reached both (the #62 class). With #66's day pin now in the RC, the chip can name its day instead of disappearing.

## The change (`state/store.ts`, the drift offer's drop chip only)
1. **The reply names its day**, in the same day words the shift chip already uses: `remove the Groceries today at 14:00`, `… tomorrow at …`, `… on thursday at …`. A block past the day words (a week or more out) keeps the day-less reply, and the guard withholds the chip if that reply could reach another day.
2. **The guard resolves the reply exactly as `execRemove` will.** The parsed `dayOffset` becomes `resolveRemoval`'s `day`. **This also fixes a latent guard bug:** it used to pass the parsed `remove` object straight through, and `resolveRemoval` ignores `dayOffset` (it reads `day`), so any day in a reply was silently dropped from the guard's prediction.

The rest of the offer is unchanged: shift, keep, a series never gets a drop chip, and nothing moves before a pick.

## Proof (`state/__tests__/drift-fallback.test.ts`, the file #63 added)
- **Updated pins** (the behaviour change):
  - The three-chip offer's drop reply now reads `remove the Groceries today at 14:00`.
  - **With a same-titled Groceries at 14:00 on Thursday**, the drop chip now shows. **Picking it removes only today's Groceries** (Thursday's stays), pinned by a full block snapshot.
  - The "neither exact" case now uses a series occurrence (which never gets a drop chip) plus a full Wednesday, since an other-day twin no longer withholds the chip.
- **New pin:** a keyed turn gets the same day-named drop chip.

**Mutants** (each applied, run, then restored):
```
M1 the drop reply stays day-less                → ✘ 3
M2 the guard ignores the day pin (the old bug)  → ✘ 2
```

## Gates (from `app/`, head `a3acee0` on RC `b1d8799`)
```
cwd=/home/work/work/mew/.worktrees/rc1-12b/app head=a3acee0
tsc exit=0
 Test Files  129 passed | 2 skipped (131)
[...]

Issue #81, opened 2026-09-17
[dev] Plan-mode pick drops a stated length: after the picker, MEW offers to pad (and pads) blocks the owner sized in their own words


Derpimort:
## What & why

**When a plan goes through the scenario picker, a length the owner stated gets padded anyway.** #11 (archive #322) made it a law: "Stated word wins: an explicit duration in the ask … is never padded." On the direct path that holds, and it's pinned. On the picker path (plan mode, or any ask that opens the picker), the stated flag is dropped between the scenario and its apply. After the pick, the estimate offer counts stated blocks as unstated, and **pad** re-sizes them.

Found while verifying #11 in the RC closure pass (coderpa). **Pre-existing** on the RC `c60c466`, not a regression from tonight's PRs.

## Evidence
Probe through the real store, keyless, on the RC `c60c466`. It uses the `estimate.test.ts` harness with deep work that runs ~20% long, `estimateAutosize: 'ask'` and `planMode: 'always'`.
```
ask: "block 90 min for the quarterly report, block 90 min for the board deck,
      block 90 min for the roadmap review, block inbox sweep, block errands"
picker shown · offers before the pick: 0
scenario places [title, durationMin, durationStated]:
  ["quarterly report",90,null] ["board deck",90,null] ["roadmap review",90,null]
  ["inbox sweep",60,null] ["errands",60,null]
after the pick: 1 offer → "your deep-work blocks tend to run ~20% long — want me to give them room?"
after "pad":    quarterly report 110 · board deck 110 · roadmap review 110
```

**Cause:**
- `store.ts` `execProposeScenarios` marks a stated length on the `ScenarioTask`: `durationStated: true` (L3925). `sizeToDemonstrated` honors it in `always` mode.
- `ScenarioPlace` (`domain/scenarios.ts`) has **no** `durationStated`, so the places the picker stores and applies don't carry it.
- `pickScenario` applies them with `execPlan(scenario.places, [])` (L5331). There `trackForEstimate(placed, p.durationStated)` (L2631) sees `undefined` and queues the stated blocks for the offer, and the pad resizes them.
- The direct path is pinned ("an explicit-duration ask is never offered on", "pad never touches a stated-duration block placed in the same turn"). The picker path has no pin.

## Suggested shape
[...]

PR #145, opened 2026-09-17
fix(plan): re-planning an existing block scores its slot and moves it with one length (#135)

### LINKED ISSUE #135: Re-planning an existing block scores the slot with the ask's length but moves the block at its own length, so it can land over the next block
## What happens
Re-planning a block that already lives on the day is a **move** (#89's same-title rule). The slot is scored with the **ask's** length, but the move keeps the **existing block's** length, so the block can land over whatever follows.

**Repro (real store, keyless floor, RC `8a1520f`):** Tuesday 14:00, with Walk (health, flexible) 9:00–10:30 and Review 14:30–22:30.
- `block 30 min for a walk` → `Done — moved walk to today 14:00–15:30 — note: it overlaps Review 14:30–22:30 (flexible).`
  - The scorer found the 30-minute gap at 14:00 (it was asked for 30 min, with the Walk itself excluded from the occupied blocks).
  - `week.move` then kept the Walk's own 90 minutes, so it runs an hour over Review.

It also shows through #117 (`block 30 min for a walk after dinner` in the demo week moves the existing 16:00 Walk to 20:00–21:00, over writing), because the title is now just "walk".

## Where
In `execPlan`'s `existing` branch (store.ts, the #89 de-dup), `SlotQuery.durationMin` is `prefd.durationMin ?? 60`, while `week.move(blocks, existing.id, key, landStart)` keeps `existing.endMin - existing.startMin`.

## Expected
One length decides both. Either the move takes the length the owner just stated (it's their word this turn), or the slot is scored with the existing block's length. Either way, a re-plan never lands over the next block through a length mismatch.

Found while live-probing #117 (coderpb).

Derpimort:
Fixes #135. Task `tsk_01m2r68tanz75msb1bmnzytz32`. Head `22802b0` on RC `131a56c`. Commits:
- `33a7d0d`: the fix
- a clean merge of the RC

## What the owner sees
```
(Walk 9:00–10:30 this morning; Review from 14:30; it's 14:00)
you ❯ block 30 min for a walk
before: Done — moved walk to today 14:00–15:30 — note: it overlaps Review 14:30–22:30 (flexible).
now:    Done — moved walk to today 14:00–14:30.
```

## Cause
Asking again for a block that already lives on the day is a **move** (#89's same-title rule). `execPlan` scored the slot with the **ask's** length (`prefd.durationMin ?? 60`), with the block itself excluded from the occupied blocks, but `week.move` kept the **block's own** length. A 30-minute gap got a 90-minute Walk.

## The change (`state/store.ts` `execPlan`)
**One `replanLen`** for an existing block: the length the owner **stated this turn** (`p.durationStated`), else the block's **own** length. It's used for:
- the scorer's query;
- the granted-overlap check;
- the meal guardrail;
- `week.move(…, replanLen)`, which already takes a length.

## Calls I made
- **A remembered length rule doesn't resize a block the owner already has.** `remember walk always takes 45 min` sizes **new** walks; `block a walk at 12:00` moves the existing 90-minute Walk and keeps its 90. A re-plan is a move, and only the owner's words this turn change its length.
- **When no slot fits the one length today,** the block stays where it is, as before.

## Proof
**Real store** (`state/__tests__/replan-one-length.test.ts`, 5):
- **the issue's repro:** `block 30 min for a walk` at 14:00 lands 14:00–14:30, with exactly the reply above, and Review is untouched;
- **no length stated:** a 90-minute Walk never lands over Review, whether the gap is 30 or 60 minutes;
- **a remembered 45-minute rule** leaves the existing Walk at 90 when re-planned at 12:00;
- **a stated length with a stated time** resizes as it moves (`block 45 min for a walk at 12:00` → 12:00–12:45).

**Mutants** (each applied; restored):
```
L1 the scorer uses the ask's length again (the bug)   ✘ 1
L2 the move keeps the block's own length              ✘ 2
L3 a remembered rule length replaces the block's own  ✘ 1
```
L1 and L3 survived my first two fixtures; the 60-minute-gap and remembered-rule cases were added to catch them.

### Live, in the built app (keyless, `?d=2026-09-16&t=14:00`, the demo week's Walk 16:00–17:30)
```
you ❯ block 2h for writing tonight
week: writing 20:30–22:30
you ❯ block 30 min for a walk after dinner
mew ❯ Done — moved walk to today 20:00–20:30.
week: Walk 20:00–20:30 · writing 20:30–22:30
```
This is the probe that surfaced #135 during #117's review: the same asks moved the Walk to **20:00–21:00, over writing**. Page errors: none. This was a throwaway Playwright probe against `pnpm preview`; it wasn't committed.

## Gates (from `app/`, head `22802b0` on RC `131a56c`, served on :5202)
```
cwd=/home/work/work/mew/.worktrees/rc1-135/app head=22802b0
tsc exit=0
 Test Files  172 passed | 2 skipped (174)
      Tests  2783 passed | 9 skipped (2792)
Lines        : 93.76% ( 7070/7540 )
coverage exit=0
lint exit=0
format exit=0
build exit=0
| ✅ | **main** (entry + its static non-vendor chunks) | | **391.1 KB** | **440.0 KB** |
| ✅ | **first load** (eager) | | **794.7 KB** | **1200.0 KB** |
bundle exit=0
shoot exit=0
overlap exit=0
```
Canon PNGs were restored after the shoot. CHANGELOG `[Unreleased]` → *Your stated lengths hold through the plan picker* gains one bullet. `check-changelog.mjs` → `31 sections (31 unique) · 57 bullets (57 unique) — no repeats`.

**Laws:**
- **The owner's stated word holds** (a length said this turn);
- **a placement never lands over the next block** through a length mismatch;
- **positive voice;**
- **tools remain the only mutation path.**


Derpimort:
**Peer review (coderpa) — APPROVE. 0 🟡, two optional 🟢.**

Read @22802b0 in its own worktree (on RC `131a56c`).

## What I checked

**One length, computed once, used everywhere.** `replanLen` is defined exactly when `existing` is — the stated length this turn (#81's `durationStated`), else the block's own — and it then feeds all four consumers: the scorer's `SlotQuery`, the meal guardrail's `correctMeal`, the granted-overlap window and `week.move`'s optional length (which `week.ts` already took). That's the bug closed at its root: the scorer can no longer pick a gap for one length while the move lands another in it.

**Nothing new for a fresh block.** With no `existing`, `replanLen` is `undefined` and every site falls back to `prefd.durationMin ?? 60`, byte-for-byte what it was. The meal line reads the same way (`existing ? own : prefd` collapses into `replanLen ?? prefd`), so the only behaviour that moved is the one the PR names.

**`replanLen!` is safe.** The assertion sits inside the `existing` branch (`landStart = start ?? existing.startMin` two lines up), and `replanLen` is non-`undefined` exactly there.

**The call is the right one.** A length said this turn resizes the owner's block; a remembered rule sizes new blocks only. `estimateSource` isn't read by any logic in the app (only written), so a resize through this path leaves nothing stale behind.

- `replan-one-length` 5 passed; with `granted-overlap`, `plan-never-past`, `plan-evening`, `scenarios`, `rc-journeys`: **352 passed**.
- Mutants (store.ts restored byte-identical after each):

| Mutant | Result |
|---|---|
| R1: a stated length never resizes the owner's block | **CAUGHT** (the repro pin + the stated-time pin) |
| R2: the scorer goes back to the resolver's length — the original bug | **CAUGHT** (the 60-min-gap pin) |
| R3: the move keeps the block's own length | **CAUGHT** (two pins) |
| R4: the meal guardrail sizes with the block's own length | **SURVIVED** (352 green) |
| R5: the granted-overlap window keeps the old length | **SURVIVED** (352 green) |

## 🟢 (optional) — the two other consumers are unpinned

Both are reachable, and in both the head is right and the mutant is wrong; only the net is missing:

- **R4, a re-planned meal with a stated length.** Existing `Lunch 12:00–13:00`, say `block 30 min for lunch at 12:30`: the head asks `correctMeal` about a 30-min window, the mutant about a 60-min one, so the guardrail can correct (or not) on a length the owner didn't ask for. A pin on a meal-class re-plan with a stated length holds it.
- **R5, a re-plan with a granted overlap and a stated length.** Existing 90-min `Walk`, say `block 30 min for the walk at 14:00, it's fine to overlap review` with a fixed block from 14:30: the head checks 14:00–14:30 and places; the mutant checks 14:00–15:30 and refuses over a block the shorter walk never touches — a false refusal in the same family #135 is about.

Both are 🟢, not 🟡: the code does the right thing today, and #135's "one length" claim would just be held by the tests as well as by the code.

Clean, well-scoped fix — the kind that deletes a class rather than a case.

PR #144, opened 2026-09-17
fix(close-loop): MEW's own seeded meals are never open items — the day clears and the evening never offers to roll them

Derpimort:
## What
Task `tsk_01m2r7cgz2f4rv23ypb2xc3mnf`: #123's class on another surface, from coderpb's #138 review. Off RC `131a56c` (the current head, so no merge is needed).

`week.openItems` and `week.dayClear` counted the meals the sustenance scaffold seeded as if they were the owner's work.

**Repro on RC `131a56c`.** Sustenance is on, MEW seeded Lunch 12:00–12:45 and Dinner 18:30–19:30, and the owner's only block (Deck 9:00–10:00) is done:
- `openItems` → `['Lunch', 'Dinner']`;
- `dayClear` → `false`, so the day never reads as clear;
- the 18:05 tick → `close-loop: Lunch isn't done — shall it live tomorrow at 9:00? Then let it go for tonight.`

MEW offered to move a meal to tomorrow that the owner never asked for, and the owner's finished day never counted as clear.

## How
`domain/week.ts`: `openItems` and `dayClear` leave out `placedBy` blocks, which is #123's marker for MEW's own scaffolding (a seeded meal, a pacing breather).

That one rule fixes every reader of the two helpers:

| Reader | Before | After |
|---|---|---|
| `nudges/engine.ts`: the close-the-loop nudge | offered to roll a seeded Lunch | offers only the owner's open work |
| `nudges/brief.ts`: the wrap debrief's `waiting` list | listed seeded meals | lists only the owner's work |
| `store.ts`: the day-end rest log (`workLeftOpen`) | a seeded meal left open logged the rest as `rest_skipped` | `rest_kept` |
| `liveNow.ts` / `pixie.ts`: the dial's resting state | never resting while a seeded meal was open | resting once the owner's work is done |

**Calls I made inside the laws:**
- **A meal the owner named is still theirs.** `lunch with sam` has no `placedBy`, so close-the-loop still offers it. Only what MEW placed on its own is left out, the same line #138 drew for the review's roll.
- **`dayEndMin` is unchanged.** It doesn't read unfinished work. It says when the working day ends, which times close-the-loop and sizes the rescue shift window (`rescue.ts`). A seeded dinner is still time the evening holds, and changing that would move both, which is outside this task.
- **Nothing is removed or moved.** The meals stay on the day exactly as placed. Only the "unfinished work" readings change.

## Tests
[...]

PR #143, opened 2026-09-17
fix(replies): a reply's question aside keeps its '?' and never gets a period after it (#126)

Derpimort:
Fixes #126. Task `tsk_01m2r4h5bgeqx48p9q69bk68zm`. Base: RC `84e93df`. One commit, `f20e10c`.

## What the owner sees
```
before: … That's your 2nd hour-plus work block this week. Today runs 17:00–20:00 unbroken — want me to make room for a short breather?.
now:    … That's your 2nd hour-plus work block this week. Today runs 17:00–20:00 unbroken — want me to make room for a short breather?
```

## Cause
`paceRest`'s wall-to-wall note already ends with a question mark. Four reply sites joined their asides with `joinHuman`, capitalised the result and appended `.` whatever it ended with:
- `execPlan`'s pacing aside;
- `execPlan`'s meal aside;
- `execSplit`'s pacing;
- the split series pacing.

A question also landed mid-sentence when it wasn't the last note (`…breather?, tucked a 15-min breather …`).

## The change (`state/store.ts`)
**`asideSentences(notes)`** replaces all four sites:
- the statements are joined into one capitalised sentence with its period, exactly as before;
- a note that already ends a sentence (`?` / `!` / `.`) stands as its own capitalised sentence after them, never given a second mark.

A single statement reads byte-identically (`Tucked a 15-min breather into today at 13:00.`).

## Proof
**Real store** (`state/__tests__/aside-punctuation.test.ts`, 3):
- **The issue's repro:** Deck 17:00–18:00 and a flexible Gym 18:00–19:00, then `block 1h for report at 18:00` → the reply ends `… unbroken — want me to make room for a short breather?`, with no `?.`.
- **A split:** `split the deck around 10:00-10:30` with Budget right after asks the same way, with no `?.`.
- **A statement aside** still ends with its period.

The full suite passed before the gates (`2776 passed`); `evening.test.ts`'s `toContain` pin is unchanged.

**Mutants** (each applied; restored):
```
P1 every aside gets a period (the old join)   ✘ 2
P2 a question joins the statements            ✘ 2
```

## Gates (from `app/`, head `f20e10c` on RC `84e93df`, served on :5202)
```
cwd=/home/work/work/mew/.worktrees/rc1-126/app head=f20e10c
tsc exit=0
 Test Files  171 passed | 2 skipped (173)
      Tests  2776 passed | 9 skipped (2785)
Lines        : 93.76% ( 7064/7534 )
coverage exit=0
lint exit=0
[...]

Issue #49, opened 2026-09-17
[dev] Let the user grant an overlap: an explicit opt-in on flexible blocks; fixed and [calendar] still refuse (split from #22)


Derpimort:
## What & why

Split from #22 (Fix 4 / AC7) per the RC plan of record, so #22's slices stay about the plannable day.

In the #22 transcript the user granted an overlap twice in plain words ("overlapping with gaming maybe?", "have the email and prod release overlap a bit with gaming"). MEW never did it, and **could not have**: consent to overlap has nowhere to go.

- No placement tool accepts an overlap grant: `plan_blocks`, `move_task`, `edit_block` (`adapters/model/tools.ts`).
- `candidateSlots` / `freeWindows` never offer an overlapping candidate (`domain/scheduler.ts`, `domain/week.ts`).
- `clashNote` writes the refusal into the tool result the model reads: `(flexible — offer to drift it, don't move it unasked)` (`state/store.ts`).
- The `<acting>` prompt backs that up: "if it overlaps a flexible block, don't silently re-place either one, just offer to drift the other side".
- The manual path bounces as well: `dragMove` returns `'conflict'`.

The data model can already represent a consented overlap: `attention: 'background'` is transparent to `conflictsWith` and to slot search. Nothing maps a spoken "overlap these" onto it.

## Scope

1. An explicit, per-call opt-in on the placement tools (e.g. `allowOverlap` on `plan_blocks` / `move_task` / `edit_block`), or a mapping of the consented side onto `attention: 'background'`. Pick one and write the reasoning into the PR.
2. When the grant is present, `clashNote` stops emitting "don't move it unasked", and the prompt's drift rule gives way to the grant for that call only.
3. **Product laws stay intact.** Fixed blocks and `[calendar]` (external) blocks still refuse: "fixed-time blocks are scheduled around, never over", and "external calendar events are never moved". The grant covers only flexible blocks the user named.
4. Offer-then-confirm stays. The grant has to come from the user's own words in the current turn. The model must never infer it.

## Acceptance criteria

- [ ] A user-granted overlap onto a **flexible** block executes, and the receipt names the overlap plainly — pinned (state test)
- [ ] The same grant onto a **fixed** block or a **[calendar]** block still refuses, with the honest reason — both pinned
- [ ] With no grant, behavior is byte-identical to today (existing collision/drift suites stay green)
- [ ] tsc · coverage · lint · format · build green

## Out of scope

[...]

PR #146, opened 2026-09-17
test(pins): six reviewer greens from tonight's rounds, each held against its own mutant (#141, #143, #144, #145)

Derpimort:
Six reviewer 🟢 pins from tonight's rounds, in one test-only PR: the optional greens coderpa and I
raised on #141, #143, #144 and #145. No product code changes — only new test cases and one
CHANGELOG line.

Each pin was written against a **mutant**: the source edited to the behaviour the review worried
about, the file re-run, the source restored byte-for-byte. A pin that passed both ways was rewritten
until it failed on the mutant. Two of them hold a mutant nothing else in the suite catches.

## The pins

| # | Review | Holds | File |
|---|---|---|---|
| Q2 | #141 | a breather MEW tucks in for pacing keeps its "(flexible)" note — `isProtectedRest` reads `protected`, not just the tag | `state/__tests__/clash-note-rest.test.ts` |
| Q3 | #141 | a rest with a standing "never moves" rule reads as fixed, because fixed-time is judged before the rest wording | same |
| M4 | #144 | a day holding nothing but MEW's own seeded meals never reads as "clear, rest earned" | `domain/__tests__/week.test.ts` |
| R4 | #145 | the meal guardrail reads the one length too: an existing 12:00–13:00 lunch asked for "30 min at 13:30" lands 13:30–14:00, inside its window | `state/__tests__/replan-one-length.test.ts` |
| R5 | #145 | a granted overlap is judged on the one length, so a 30-min walk at 14:00 clears a call fixed at 14:30 | same |
| Q1 | #143 | with two days planned, the statement sentence comes first and the question stands last, with "?" once | `state/__tests__/aside-punctuation.test.ts` |

## Mutants, and which pin fails on each

#145's four mutable spots, each run alone against `replan-one-length.test.ts`:

| Mutant | Tests failing | Caught by |
|---|---|---|
| score the slot with the preferred length | 1 | #145's own test |
| land with the block's own length | 1 | **R4 only** |
| take the granted-overlap window from the block's own length | 1 | **R5 only** |
| move without the one length | 4 | #145's tests + R4 + R5 |

The other three:

- `dayClear` counts MEW's seeded meals as items → **M4** fails (1 of 88).
- `dayClear` drops its `length > 0` guard, so a day of nothing but scaffolding reads clear → **M4** fails (1 of 88).
- the asides put question notes before statements → **Q1** fails (1 of 4).

Q2/Q3 were verified the same way in the earlier round (coderpa pre-ran the file at `3de4310`: 7 passed).

## Two calls I made, written down

- **R5 goes through the keyed model.** `allowOverlap` never reaches `execPlan` from the keyless
  floor — "it's fine to overlap the call" parses to a plain placement with no grant, and the em-dash
  variant folds the clause into the title. Granted overlap is a connected-model input (#49), so the
  pin drives the scripted local model, exactly as `clash-note-rest.test.ts` does for its keyed
  results. Nothing touches the network.
- **Q1 goes through the keyed model for the same kind of reason.** One keyless command cannot
  straddle two days — in "…at 17:30 and 1h for the review tomorrow at 11:00", `tomorrow` binds both
  specs and everything lands on one day, which raises only one note. A two-day plan is a
  `exec.plan([...])` with two `dayOffset`s.

Both are noted in the test files themselves, next to the fixture.

## Gates — merged head `1691b95` (RC `c1154af` merged in, #145 included)

```
tsc exit=0
coverage exit=0      Test Files  175 passed | 2 skipped (177)
                           Tests  2804 passed | 9 skipped (2813)
                     All files | 91.67 | 83.71 | 91.07 | 93.78
lint exit=0
format exit=0
build exit=0
bundle exit=0        main       392.9 KB / 440.0 KB
                     first load 796.5 KB / 1200.0 KB
                     total     1545.9 KB / 2300.0 KB
```

Changelog guard: `[Unreleased]` 31 sections (31 unique) · 61 bullets (61 unique) — no repeats. The
CHANGELOG line was hand-edited, never prettier, per the standing rule.

Shoot and overlap are not required: this PR touches `CHANGELOG.md` and four test files, no `ui/`,
no `scripts/`, no product code. Working tree clean after the chain.


Derpimort:

Read @b184d29 in its own worktree (RC `4063643` already merged in, MERGEABLE). Test-only: 184 insertions, **zero deletions**, no `src` change outside tests, plus one CHANGELOG bullet.

## The thing that matters in a test-only PR: does each pin bite?

I re-ran all six mutants myself on your head — not your table, my own runs — against the four touched suites. **Every one is caught, and each by exactly one test:**

| Mutant | Result | The single test that failed |
|---|---|---|
| `isProtectedRest` drops `protected` | CAUGHT | Q2 — a pacing breather keeps "(flexible)" |
| the rests filter drops `!isFixedTime` | CAUGHT | Q3 — a rest ruled "never moves" reads fixed |
| `dayClear` drops `day.length > 0` | CAUGHT | M4 — a day of nothing but MEW's meals isn't clear |
| the meal guardrail sizes with the block's own length | CAUGHT | R4 — 13:30 fits the lunch window at 30 min |
| the granted-overlap window takes the block's own length | CAUGHT | R5 — the 30-min walk clears the 14:30 call |
| `asideSentences` emits questions before statements | CAUGHT | Q1 — statement first, question last |

106 passed across `week`, `aside-punctuation`, `clash-note-rest` and `replan-one-length` on the clean head; sources restored byte-identical after each mutant. (I verified uniqueness *within the four touched suites* — your table's wider claim across the full suite I did not re-run, since it wouldn't change the verdict.)

## Your two calls — both right, and one of them corrects me

- **`allowOverlap` is model-only (#49).** I checked: the identifier appears in `adapters/model/tools.ts`, `adapters/model/types.ts` and `state/store.ts`, and **nowhere in `domain/parse.ts`** — the keyless grammar cannot set it. So the keyless repro I suggested in my #145 review ("block 30 min for the walk at 14:00, it's fine to overlap review") would never have granted the overlap. Driving the scripted keyed model is the correct route, and the fixture says so beside itself. Good catch on my review, not just on the code.
- **One keyless command can't straddle two days**, so Q1 goes through the model too. The fixture then does something better than my suggestion: the question note is raised for *today* while the breather is tucked into *Wednesday*, so the notes arrive in question-then-statement order and the pin proves the reply **reorders** them. That's the real claim — a pin that merely observed the incidental order would have been worth nothing.

## The pins themselves

- **M4** is three assertions, not one: all-scaffolding → not clear; with the owner's own block done → clear even with the seeded meals still open; with it open → not clear. That isolates the length guard exactly.
- **R4** asserts the whole reply, not a fragment: `Done — moved lunch to today 13:30–14:00.` Under the mutant the 60-min lunch runs past the window and the aside changes — which is the behaviour, not the wording, doing the work.
- **R5** pairs a negative (`not.toContain('stays unplaced')`) with the positive placement `[[TODAY, 14:00, 14:30]]`, so it can't pass vacuously.

CHANGELOG: one bullet, hand-edited, under "Under the hood". Guard 31 sections / 61 bullets, no repeats.

Nothing to change. This is the closing brace on tonight's review loop — six greens raised across four PRs, all six now held by tests that fail without the code.

PR #138, opened 2026-09-17
fix(review): the weekly review carries only the owner's work — MEW's own breathers and meals never roll (#123)

Derpimort:
## What
Fixes #123 (task `tsk_01m2r3yzj2jnw652m874vv26v1`). Off RC `36eebab`.

The weekly review's **carried** list (`isRollCandidate`: any open, own, flexible block) offered **MEW's own scaffolding** to roll into next week: the `Breather` the pacing pass tucks in, and the `Lunch` and `Dinner` the sustenance scaffold places (tally `rest · 1 to carry`). Found by the RC journeys (#125, journey 4).

## How
- **`domain/types.ts`:** `Block.placedBy?: 'pacing' | 'sustenance'` marks MEW's own scaffolding. It's unset on every block the owner asked for, **including a meal or rest they named**.
- **`domain/week.ts`:** `PlaceSpec.placedBy` flows into the placed block.
- **`state/store.ts`:**
  - `paceRest` places its breather with `placedBy: 'pacing'`.
  - The sustenance pass marks the meals it just placed (`'sustenance'`), plus any breather (`'pacing'`), right after its plan runs.
- **`domain/review.ts`:** `isRollCandidate` admits no `placedBy` block. That one predicate gates both the review's list and `rollForward` itself, so a scaffolding id handed straight to `rollForward` rolls nothing.

**Calls I made inside the laws:**
- **A provenance mark, not a title rule.** The scaffold's meals are titled plain `Lunch` and `Dinner`, so a title rule would also have kept an owner's own lunch from carrying. The manager's brief asks to keep those rollable. `lunch with sam`, typed by the owner, still carries.
- **Legacy blocks carry no mark.** Scaffolding placed before this change (the current week, at most) can still show once. Everything placed from now on is marked, and next week's scaffolding places its own.

## Tests
New `state/__tests__/review-carries-owner-work.test.ts` (4, real store):
1. **A split's pacing breather** is marked `pacing` and not carried. The split's pieces and the Wednesday draft are: `['Roadmap draft', 'Spec review', 'Spec review (part 2)']`.
2. **The sustenance scaffold's meals** are marked `sustenance` and not carried, while `lunch with sam` (typed) is.
3. **Handing the breather's id to `rollForward`** rolls nothing: it stays open, nothing appears next week, and no chat is posted.
4. **The mark is stored** (`fakeDb`), and after a reload the breather still isn't carried.

**Existing pins touched:**
- The journeys' **#123 pin flips** to a plain `it` and passes.
[...]

Issue #123, opened 2026-09-17
[dev] The weekly review offers MEW's own Breather (and seeded meals) as carried work to roll into next week


Derpimort:
## What happens
The weekly review offers MEW's own **Breather** as carried work to roll into next week.

**Repro (real store, keyless floor):** Friday 10:00. Roadmap draft on Wednesday; Spec review 13:00–15:00 Friday; a calendar Client call 14:00–14:30.
1. `split the spec review around the 2pm call` → `Split — Spec review now runs 13:00–14:00, and Spec review (part 2) picks up 14:30–15:30, around Client call 14:00–14:30. Tucked a 15-min breather into today at 15:30.`
2. `openWeeklyReview().carried` → `["Roadmap draft", "Spec review", "Spec review (part 2)", "Breather", …]`

The Breather MEW tucked in for pacing (tag rest) appears among the blocks the owner is invited to roll forward, and the tag tally reads "rest · 1 to carry". The roll-forward tests already note that seeded meals are carried the same way. `isRollCandidate` (domain/review.ts) admits any open, own, flexible block, whatever it is.

## Expected
"Carried" is unfinished work the owner might roll. MEW's own pacing breathers, and the meals the sustenance scaffold placed, aren't offered to roll into next week. Next week's scaffold and pacing place their own.

Found by the RC cross-feature journeys (`rc-journeys.test.ts`, test-only). It's pinned there with `it.fails`.

PR #147, opened 2026-09-17
fix(rail): the loose-threads rail holds the owner's own work, never MEW's scaffolding

Derpimort:
## What
Task `tsk_01m2ra2jw9wdak7hmzbfpjdt2r`, found by auditing #123's class after #144. Off RC `131a56c`, with the whole train (`4063643`) merged in and re-gated on that head.

The loose-threads rail listed **MEW's own scaffolding** as the owner's slipped work. `week.looseThreads`' `slipped` list took every open, non-background, non-optional, non-all-day block whose window had passed — with no `placedBy` filter — and `ThreadRail` renders each row as `was 12:00–12:45` with **done** and **resume** beside it.

**Repro on RC `131a56c`** (real store, sustenance on, Tuesday; the owner's Deck 9:00–10:00 with MEW's seeded Lunch 12:00–12:45 and Dinner 18:30–19:30). At 15:00:

```
RC        slipped: ["Deck 540-600", "Lunch 720-765"]
this PR   slipped: ["Deck 540-600"]
```

So MEW asked the owner to tick off — or pick back up — a lunch MEW placed for them.

## How
`domain/week.ts`: `slipped` leaves out `placedBy` blocks, the same one rule #138's roll and #144's open items draw. A seeded meal and a tucked-in breather are MEW's scaffolding, never the owner's loose ends.

**The other three lists were checked and needed nothing:**

| List | Why it's already right |
|---|---|
| `running` | background blocks only; scaffolding is never background |
| `paused` | derives from `rolledToId`, and #138 keeps scaffolding out of the roll, so it's never a roll target |
| `unplaced` | captures, not blocks |

**Calls I made inside the laws:**
- **The line is who placed it, not what it is.** `lunch with sam` has no `placedBy`, so it still slips and still gets its row — only what MEW placed on its own steps out.
- **Nothing is removed or hidden from the week.** The seeded meal stays exactly where it is on the dial and in Week; only the "still loose" reading changes.
- **`insights.ts`'s "slipped" line is not in this class** — it reads completion events (blocks the owner ticked late), not open blocks. Left alone.

## Tests
New `domain/__tests__/loose-threads-scaffold.test.ts` (3, pure — the rule lives in the domain and `ThreadRail` renders `looseThreads` and nothing else):
1. **The repro:** with a seeded Lunch, a pacing Breather, a seeded Dinner and the owner's Deck on the day, `slipped` at 15:00 is `['Deck']`.
2. **An owner-named meal is still theirs:** `lunch with sam` slips, MEW's Lunch doesn't.
3. **The other three lists are exactly as they were** with scaffolding in the day: `running` `['r1']`, `paused` `['p1']`, `unplaced` `['c1']`, `slipped` `[]`.

**Mutants (all caught):**

| Mutant | Caught by |
|---|---|
| N1: the rail keeps MEW's scaffolding (**the RC's own behaviour**) | 1, 2, 3 |
| N2: the rule inverted (only scaffolding slips) | 1, 2, 3 + the existing `looseThreads` pin |
| N3: the line drawn on the tag (`private`/`rest`) instead of `placedBy` | 2 |

`week.ts` was restored and verified byte-identical after each.

**CHANGELOG:** its own section, placed deliberately *before* "The weekly review's roll really moves your work" so it can't collide with #144's bullet at that section's end. Its own section, placed before "The weekly review's roll really moves your work" so it could not collide with #144's bullet at that section's end — which held: the RC merge was clean. Guard on this head: **32 sections / 62 bullets, no repeats**.

## Gates
```
head 410a9d1 (fix/loose-threads-scaffold with RC 4063643 merged in)
install exit 0
tsc exit 0
lint exit 0
format exit 0
coverage exit 0
build exit 0
bundle exit 0
shoot exit 0
overlap exit 0
tree: 0 change(s)
 Test Files  178 passed | 2 skipped (180)
      Tests  2832 passed | 4 skipped (2836)
Statements   : 91.62% ( 8415/9184 )
Branches     : 83.64% ( 6776/8101 )
Functions    : 91.04% ( 1729/1899 )
Lines        : 93.77% ( 7156/7631 )
| ✅ | **main** (entry + its static non-vendor chunks) | | **394.8 KB** | **440.0 KB** |
| ✅ | **first load** (eager) | | **798.4 KB** | **1200.0 KB** |
| ✅ | **total** | | **1548.2 KB** | **2300.0 KB** |
check-bundle-size: OK
done → /home/work/work/mew/.worktrees/rc1-loose/app/shots
```


Derpimort:
## Peer review (coderpb): `410a9d1`

**Verdict: APPROVE. Zero 🟡.** One optional 🟢 for later, and two non-findings written down so the class sweep is on the record.

The fix is the right shape and the right discriminator: `!b.placedBy` matches `dayClear`, `openItems` (#144) and `isRollCandidate` (#123), so all four now draw the same line — who placed it, not what it's tagged. The doc comments were updated with the code, which is why the next reader will know the rail is the owner's own loose ends.

### I re-ran your mutants myself, and added two

`week.ts` restored byte-identical after each; `loose-threads-scaffold.test.ts` + `week.test.ts` (90 tests) each run.

| mutant | tests failing |
|---|---|
| N1 the RC's own behaviour (drop the filter) | 3 |
| N2 inverted — only MEW's scaffolding slips | 4 |
| N3 drawn on the tag, not on who placed it | 3 |
| **P1 (mine)** only seeded meals excluded, a pacing breather still slips | 2 |

P1 is the one I most wanted to see fail, because a filter written as `placedBy !== 'sustenance'` would have passed a meals-only reading of the bug report. Your repro case covers the breather explicitly, so it fails. Good.

### The other three lists: verified by mechanism, not by assertion

Your note says they need nothing, and they don't — but I wanted the reason in code rather than in prose, because that's what tells us whether a future change reopens the surface:

- **`running`** filters `isBackground(b) && startedAt != null`. Scaffolding is never background, so it cannot appear. Structural.
- **`paused`** holds blocks that a *rolled* block points at. Scaffolding can never become one, because `isRollCandidate` already carries `!b.placedBy` (#123) **and** `rollForward` re-checks that gate on the ids handed to it, refusing a stray. So the protection is enforced at both ends, not incidental.
- **`unplaced`** is captures — a different type entirely.

**Optional 🟢, not blocking, not for this PR:** the `paused` surface is safe *because* of `isRollCandidate`'s `!b.placedBy`. Nothing currently pins the two together — a test that a seeded meal handed straight to `rollForward` never becomes a paused thread would hold that link, so a future edit to the roll gate can't quietly re-open this rail. Worth a line on the pins list rather than widening this PR.

### Two non-findings, on the record

- `daySummary` / `daySummaryLine` (dialDay.ts) count **every** block on the day, scaffolding included — "3 blocks · 2h committed". I checked it deliberately and it is **not** this class: the line is a positive count by construction and never frames anything as unfinished, missed or slipped. A seeded lunch really is on the day. Out of scope, correctly.
- `insights`' slipped line reads completion events rather than open blocks, as you said — I confirmed it doesn't touch this list.

### One thing for the sequence, not a change request

Your new `### The loose-threads rail holds your own work` section has no blank line after the heading, so it reads in the **hand** style rather than the file's current prettier shape. That is consistent with the restoration you're about to land, and the guard is clean (32 sections / 62 bullets), so I'd leave it exactly as it is — flagging it only so nobody "tidies" it into prettier's shape in between.

Merging next: the RC moved under you (#146 landed at `39ebfd8`), so I'll re-sync, hand-verify the CHANGELOG, run my own chain, pin the CI poll to the SHA I push, then squash.

Issue #149, opened 2026-09-17
[dev] Undo after a retag says the blocks were put "back where they were" — a retag moves nothing


Derpimort:
Found by cross-feature journeys round 2 (`rc-journeys-2.test.ts`), where #130's undo meets #142's retag. Test-only PR; no product code changed.

## What happens

A retag moves nothing — that's its whole point ("a retag keeps every block where it is"). But `execUndo`'s summary describes every *changed* block with the move vocabulary, so undoing a retag tells the owner their blocks were put back **where they were**, which is not what changed.

```
you   tag all of tomorrow's calls as work
MEW   tag 2 blocks as work tomorrow? Client call 9:00 · Vendor call 11:00.
      Board call 15:00 (from your calendar) and Standup call 17:00 (repeats) keep their tags.
you   [do it]
MEW   Tagged 2 blocks as work tomorrow — Client call 9:00 · Vendor call 11:00. …
you   undo that
MEW   Undone — put two blocks back where they were.      ← they never went anywhere
```

The week itself is correct: both tags return, times untouched, the calendar event and the repeating block are never touched at any step. Only the sentence is wrong.

## Where

`state/store.ts`, `execUndo`'s summary (~`store.ts:5641`):

```ts
if (changed.length)
  parts.push(
    `put ${…} back where ${changed.length === 1 ? 'it' : 'they'} ${changed.length === 1 ? 'was' : 'were'}`
  )
```

`changed` is "a block that exists in both snapshots but differs" — which since #142 includes a block whose only difference is its tag.

## Suggested shape (not in this PR)

Split `changed` by what actually differs: a block whose day/time moved keeps "put … back where it was"; a block whose tag changed reads in the tag's own words ("put both tags back", "put Client call's tag back"); a block that changed both can keep the time wording, which is the larger fact. Same one-sentence voice, and it keeps working when the next batch op lands (#75 slice 3).

## Pin

`rc-journeys-2.test.ts`, RC journey 7, step 4 — `it.fails` until this is fixed, asserting the line does **not** say "back where they were" for a retag-only undo. The rest of that journey (the offer, the skips, the tags, the restore) passes and stays passing.

Severity: cosmetic, but it's a receipt that misdescribes what MEW did, which is the one thing a receipt is for.

Derpimort:
## The class is wider than a retag: a resize and a rename say it too

[...]

PR #150, opened 2026-09-17
feat(batch): a sweep over a repeating block asks which occurrences you mean, and nothing moves until you answer (#75 slice 3)

Derpimort:
#75's last slice: a batch that reaches a **repeating** block asks which occurrences you mean, with the same three answers a single series edit offers (#343) — and nothing moves until you answer.

Until now a sweep named every repeating block and left it alone, which is safe but a dead end: there was no way to say "push my Tuesday-through-Thursday gym back half an hour" at all.

## What it does

"push all health after 4pm today later by 30 min", with a Gym that repeats Tue/Wed/Thu:

> Gym repeats — which do you mean?
> `just this one` · `this & the ones after` · `the whole series`

Each chip re-issues the sweep **in the keyless batch grammar** with its scope word — `push all health after 16:00 today later by 30 min across the whole series` — so a pick means the same list on either floor, and it rides the same midnight re-check every other chip gets (#94/#96). Answering widens the list, so the ordinary wide-change confirm still shows exactly what moves before it moves.

## Four calls I made, and why

**1. MEW asks first and moves nothing** — not even the one-off blocks in the same sweep that it could safely have moved. That is what makes a pick exact: every chip re-issues the *same* ask, so a sweep that had already shifted its one-offs would shift them a second time when the answer came back. Asking first is also simply offer-then-confirm: the ask as given was ambiguous.

**2. Your answer travels with the yes.** A wide change re-asks the sweep when you confirm, and without the scope word on that re-ask MEW raised the series question *again* instead of acting. My own store tests caught it. The scope phrase now sits before the `— yes, all N · TOKEN` tail: one is lifted off the front before the grammar reads the ask, the other read off the end, so the two never collide (pinned in the domain suite).

**3. A move onto one day never asks**, because two of the three answers cannot act on it — stacking a whole run onto Thursday is never what "move them" means. Rather than offer two dead chips, the reason line says what does work: `Gym 18:00 (repeats — "just this one" moves this one)`. A `series-day` skip covers a keyed model that asks for it anyway.

**4. Fixed-time is read before the series question**, so "the whole series" can never become a way around it: a repeating standup stays put for a move and still takes a retag, exactly like a one-off standup. A series holding a single open occurrence asks nothing either — the three answers would collapse into one — so it stays named and where it is, as today.

## One vocabulary, one day naming

The canonical re-ask the yes chip already used is now a single builder both chips call, so every chip names its day (never positionally) and a pick means the same list on either floor.

## Tests: 28, and what each mutant proves

`batch-slice3.test.ts` (19, domain + grammar + chip re-check) and `batch-series-scope.test.ts` (8, the real store), plus one invariant test.

| mutant | tests failing |
|---|---|
| "after" ignores the clock inside a day | 1 |
| "following" behaves like the whole series | 2 |
| a series moves with no answer at all | 2 |
| a whole series collapses onto one day | 1 |
| "just this one" narrows the sweep instead of the series | 1 |
| a scope word never reaches a batch | 2 |
| the ask is raised for a move onto one day (two dead chips) | 1 |
| the ask is raised for a one-occurrence series | 1 |
| the yes drops the answer, so MEW asks twice | 2 |
| a chip re-issues the wrong answer | 2 |

**One surviving mutant, declared rather than hidden:** the store's `scope` guard on the ask is unreachable, because a plan given an answer reports no `repeating` skip at all. I did not delete the guard (a future domain change would turn an unanswerable loop loose on the user) and I did not pretend a test covers it. Instead the invariant it leans on is pinned — *every* scope × *every* op over a pool holding a calendar occurrence, a done one, a fixed-time one and a second series — and the guard's comment says it is defensive. If the domain ever changes, that test fails before the guard matters.

Two things my own tests caught while writing them, both recorded because a green suite that never could have failed is worth nothing: a `Standup` fixture was being skipped as **fixed**, not `repeating`, so the scope code was not being exercised at all (`standup` is a FIXED_WORD); and the first `just this one` case passed its mutant because the answer must narrow the *series*, never the selection — a sweep with two one-offs and one occurrence still moves all three.

## Also

- The keyed floor takes `scope` on `batch_blocks` (`this` / `following` / `series`), described so a model knows a `move_to_day` only accepts `this`; the tool description no longer claims repeating blocks never move.
- `batch-blocks.test.ts` pinned `exec.batch`'s exact argument list, so that arity expectation is updated — flagged because it is an existing file, not one of mine.
- CHANGELOG: written by hand into the restored file, and the two `[Unreleased]` bullets slices 1 and 2 left behind are **amended**, because they said repeating blocks "keep their place" and "keep their tags" and this makes that untrue. Both were unreleased, so it is the same bullet telling the truth, not a rewrite of shipped notes. Guard: 32 sections (32 unique) / 64 bullets (64 unique), no repeats.

## Gates — RC `b2ed943` merged in

```
tsc exit=0
coverage exit=0      Test Files  180 passed | 2 skipped (182)
                           Tests  2866 passed | 4 skipped (2870)
                     All files | 91.67 | 83.74 | 91.12 | 93.70
lint exit=0
format exit=0
build exit=0
bundle exit=0        main       395.8 KB / 440.0 KB
                     first load 799.4 KB / 1200.0 KB
                     total     1549.8 KB / 2300.0 KB
```

Precisely: the chain ran on the merge head `93e2b43` with the CHANGELOG edit in the working tree; the pushed head `1a85ba7` commits exactly that edit and no code change (and the root CHANGELOG is outside `app/`'s gates by design — see RELEASES.md). Working tree clean after the chain.

Shoot and overlap are not required: `app/src` and `CHANGELOG.md` only, no `ui/`, no `scripts/`. The chips this adds are posted by the existing choices surface, which the overlap proof already covers — and it ran green on this RC tip in my #147 chain.


Derpimort:
**Peer review (coderpa) — one 🟡, then APPROVE.** The behaviour is right everywhere I pushed it; what needs a fix is the sentence the owner says yes to.

Read @1a85ba7 in its own worktree (RC `b2ed943` merged in).

## 🟡 The confirm can't be read once a scope crosses days

This is the first time a batch plan's list spans more than one day — your own comment says so ("the plan's list can now cross days, and `batchToken` follows it"). The token followed; the copy didn't. Probed on your head, `push all health after 4pm today later by 30 min` → **the whole series**:

```
OFFER:   move 4 blocks 30 min later today? Deck 17:00→17:30 · Gym 18:00→18:30 · Gym 18:00→18:30 · Gym 18:00→18:30.
RECEIPT: Moved 4 blocks 30 min later today — Deck 17:00→17:30 · Gym 18:00→18:30 · Gym 18:00→18:30 · Gym 18:00→18:30.
WEEK:    Gym Tue 18:30 · Gym Wed 18:30 · Gym Thu 18:30 · Deck Tue 17:30   ← the week is exactly right
```

Two things are wrong in the words, both in the message the batch law asks the owner to approve ("a yes acts only on the list it was offered"):

1. **"today"** covers a change that moved Wednesday's and Thursday's Gym too. `change` still reads `…${onDay(sel.dayKey)}`, the selector's single day.
2. **Three identical rows.** `line()` is `title start→end` with no day, so the three occurrences are indistinguishable — the owner can't tell what they're agreeing to, and the receipt afterwards can't be checked against the week.

Everything underneath is correct: the right four blocks move, each on its own day, and the token pins that exact list. It's only the sentence.

**Suggested shape** (yours to choose): when `plan.moves` span more than one `dayKey`, name the day on each line (`Gym Wednesday 18:00→18:30`) and let `change` drop the day claim (`30 min later`, with the days carried by the lines). Single-day sweeps — every batch before this PR — keep exactly the wording they have now, so nothing else moves.

## What I checked, and what holds

**The design calls are right, and the ordering is what makes them safe.** External → done → fixed-time → the series question → status: fixed-time is genuinely read before the scope, so "the whole series" can't become a way around it. `withScope` reads "after" day-first then clock, so an earlier occurrence on the same day is not "after", and `seriesOf` being open-only keeps a done occurrence out. The ask fires only for a series with more than one open occurrence, and never for a move onto one day — where the reason says what *does* work rather than offering two dead chips.

**Asking before moving anything is the right call**, for exactly the reason in your comment: each chip re-issues the same sweep, so a half-done sweep would shift its one-offs twice. Your tests pin that the one-off Deck sits still until the answer.

**Two asks, not one.** The scope question and the wide-change confirm stay separate, each a single calm question. That's offer-then-confirm twice rather than a fused mega-prompt, and it's what lets the token still mean the widened list.

**`batchAsk` is the right refactor:** one grammar builder now feeds both the scope chips and the yes, so the two can't drift; the scope phrase goes in front and the yes tail on the end, which is why they don't collide.

- `batch-slice3` + `batch-series-scope` + `batch-slice2` + `batch-blocks` + `batch-slice2-store`: **79 passed**.
- My own mutants (sources restored byte-identical after each) — **all five caught**:

| Mutant | Result |
|---|---|
| V1: "after" read on the clock only, not day-first | CAUGHT (2 pins) |
| V2: a series with one open occurrence still asks | CAUGHT |
| V3: the yes doesn't carry the scope word | CAUGHT (2 pins) |
| V4: a move onto one day is no longer refused for a series | CAUGHT |
| V5: the scope narrows the selection instead of widening the series | CAUGHT (4 pins) |

**On the surviving mutant you declared equivalent:** agreed, and the way you handled it is right — the `scope` guard in `askable` is unreachable because a plan given an answer reports no `repeating` skip, you pinned that invariant instead of deleting the guard, and you left a comment saying it's defensive. That's the honest handling.

**The amended slice 1/2 bullets** ("repeating blocks keep their place / their tags") are right to amend: unreleased, and this slice makes them untrue.

Fix the 🟡 and I'll merge it — everything else here is ready.

Derpimort:
## 🟡 fixed — `cb274e1` (RC `15a8fc2` merged in)

You were right, and right about where it mattered: the week was fine, but the sentence the batch law exists to get right — the one the owner says yes to — was describing a different change.

**What it says now**, same repro (`push all health after 4pm today later by 30 min` → the whole series):

```
move 4 blocks 30 min later? Deck today 17:00→17:30 · Gym today 18:00→18:30 ·
Gym tomorrow 18:00→18:30 · Gym Thursday 18:00→18:30.
```

and the receipt matches it word for word (`Moved 4 blocks 30 min later — …`). Both are now pinned as exact strings, not a `toMatch` on the front of the sentence — which is how the old wording slipped past my own tests in the first place.

I took your suggested shape as given: when the plan touches more than one day, each row carries its own day and `change` drops the day claim it can no longer make. Single-day sweeps are byte-identical — and well fenced: forcing `spansDays` true breaks **18** tests across the batch suites and the journeys.

### One thing I fixed that you didn't ask for

The other half of the same sentence had the identical ambiguity. A row that **stays put** had no day either, so a fixed call keeping Wednesday's occurrence read as:

> `Gym 18:00 (would sit over Client call 18:30–19:00) stays where it is.`

— "18:00" being the day the sweep started from. Both halves name their day now, or neither does. New test for it, and the offer reads:

> `… Gym tomorrow 18:00 (would sit over Client call 18:30–19:00) stays where it is.`

A move onto one day is deliberately excluded: its own wording already names both days and every row lands on the target, so it is untouched.

### Mutants on the fix

Each run alone against `batch-series-scope`, `batch-blocks`, `batch-slice2-store` and `rc-journeys-2`; source restored byte-for-byte after each:

| mutant | tests failing |
|---|---|
| the day claim and the rows go back to how they were | 2 |
| every sweep names days, single-day ones included | **18** |
| the rows stop naming their day | 2 |
| a move onto one day treated as a crossing plan | 1 |

### Gates on `cb274e1`

```
tsc exit=0
coverage exit=0      Test Files  181 passed | 2 skipped (183)
                           Tests  2871 passed | 1 expected fail | 4 skipped (2876)
                     All files | 91.67 | 83.76 | 91.14 | 93.70
                     main 396.0 KB / 440.0 KB · first load 799.6 KB / 1200.0 KB
```

The one expected fail is your #149 pin, now in the RC. CHANGELOG: only my slice-3 bullet plus the two amendments differ from the RC; guard 32 sections (32 unique) / 65 bullets (65 unique), no repeats. Tree clean.

### And a refinement to the stale-head rule, from this very push

Pinning to the pushed SHA was not enough on its own. My first poll read `headRefOid` **equal** to what I pushed and `pending` = **0** — because the workflows had not been *created* yet, so there was nothing pending to count. `mergeStateStatus` was `UNSTABLE`, which is what gave it away. I now poll until the **expected check set is present and passing** (all six: bundle, lint, lint-full, release-guard, typecheck, unit), which went `passing=1 pending=5` → `4/2` → all six green, `CLEAN`. Worth adding to the rule: *the SHA matching plus nothing pending can both be true while CI has not started.*

Thanks for the five mutants you ran — that's the second review tonight where you re-derived the evidence instead of reading my table.

Derpimort:
**Re-review (coderpa) — 🟡 RESOLVED. APPROVE, with one optional 🟢.**

Read the delta `1a85ba7...cb274e1` (RC `15a8fc2` merged in). The fix is exactly the scoped shape and nothing wider — `spansDays`, `onDayOf`, `rowDay`, and both the offer and the receipt reading them through the same `line` / `change` / `stays` builders, so they can't drift.

## The 🟡 is closed, checked against the acceptance

| Acceptance | Verified |
|---|---|
| rows name their day when the list crosses days | yes — and the stays line does too, which wasn't asked for and is right |
| the "today" claim is dropped when it can't be made | yes, via `onDayOf` |
| single-day wording byte-identical | **yes, and heavily pinned** — forcing `spansDays` true on every sweep fails **16** existing tests |
| the multi-day case pinned | yes — the four-row series, exact string, offer and receipt |

My own mutants on your head (`store.ts` restored byte-identical after each):

| Mutant | Result |
|---|---|
| X1: `spansDays = false` — the bug, restored | **CAUGHT** (the whole-series pin + the new stays-row pin) |
| X2: `spansDays` true for every sweep — single-day regression | **CAUGHT**, 16 fail |
| X3: a move onto one day treated as crossing | **CAUGHT** |
| X4: `spansDays` reads only `plan.moves`, ignoring `plan.skipped` | **SURVIVED** |

Suites on your head: `batch-series-scope` + `batch-slice3` + `batch-blocks` + `batch-slice2-store` + `rc-journeys-2` → **65 passed, 1 expected fail** (that one is my #149 pin now in the RC, not yours).

## 🟢 (optional) — the half of your own fix that nothing holds

You deliberately counted the **skipped** blocks' days as well as the moves', and you were right to: a sweep can move blocks on one day while naming stays on others. Nothing pins that half, so X4 survives.

Reachable shape: a Tue/Wed/Thu series, answered "the whole series", `+30 min`, with a calendar event at 18:30 on **both** Wednesday and Thursday. Those two occurrences skip as *lands-on*, only Tuesday's moves — so `plan.moves` is single-day while `plan.skipped` crosses days. Today that reads `Gym Wednesday 18:00 (would sit over …) and Gym Thursday 18:00 (…)`; with X4 it goes back to two identical rows under a "today" that doesn't cover them.

Code's right, net's missing. It belongs in a pins round or in my journeys round 3 — say which you'd prefer and I'll take it there rather than hold this PR for it.

Merging now under my own chain.

PR #151, opened 2026-09-17
test(threads): MEW's own scaffolding can never become a paused thread, and a finished follow-up leaves the rail (#147 review)

Derpimort:
The optional 🟢 I raised while reviewing #147, now held by tests — plus one gap the mutants found on the same surface.

#147 closed the rail's **slipped** list: a lunch or breather MEW placed for you no longer shows as something that slipped. While reviewing it I checked the other three lists in code rather than taking the PR's word, and `paused` turned out to be safe for a reason **nothing held**:

- `paused` lists open blocks that some **rolled** block points at (`rolledToId`).
- Scaffolding can never be rolled, because `isRollCandidate` carries `!b.placedBy` (#123) **and** `rollForward` re-checks that gate on the ids handed to it.

Two independent ends, and no test tied either of them to the rail. An edit to the roll gate would have quietly re-opened the surface #147 had just closed, with nothing failing.

## The five cases

The **control comes first**, deliberately: the owner's own block, rolled, *is* a paused thread. Without it, an empty `paused` list in the other cases would prove nothing at all.

1. the mechanism is live — own block rolled ⇒ its copy is the paused thread
2. a seeded meal **and** a breather handed straight to `rollForward` refuse: still open, no copy, nothing points at them, rail empty
3. a mixed pick (`['deck','lunch','breather']`) rolls only the owner's own work, and only that shows as paused
4. a follow-up you have finished leaves the rail — `paused` is open work only
5. the review never offers the scaffolding, so such an id could only ever arrive by mistake in the first place

## Mutants

Each run alone, source restored byte-for-byte after each:

| mutant | tests failing |
|---|---|
| the roll gate drops `!b.placedBy` | 3 of 5 |
| only seeded meals excluded, so a **breather** still rolls | 3 |
| `rollForward` stops re-checking the gate and trusts the ids it is handed | 2 |
| the rail's `paused` list stops requiring `open` | 1 |

The third is the "enforced at both ends" claim from my review — now a test rather than a sentence. The second is the same meals-only misreading I mutated coderpa's fix with; a breather is not an afterthought here.

**Case 4 came from a mutant, not from the review.** I mutated the *rail* as well as the gate, and nothing across the three loose-threads suites — 110 tests — required `paused` to hold **open** work. So a follow-up you had already ticked off would have kept sitting in the rail as a loose end: #123's class, same surface, previously unheld. It is pinned here rather than filed, because it is one line of fixture in a file that was already about this rail.

## Gates — RC `b2ed943`

```
tsc exit=0
coverage exit=0      Test Files  179 passed | 2 skipped (181)
                           Tests  2838 passed | 9 skipped (2847)
                     All files | 91.64 | 83.66 | 91.10 | 93.70
lint exit=0
format exit=0
build exit=0
bundle exit=0        main       394.8 KB / 440.0 KB
                     first load 798.4 KB / 1200.0 KB
```

Changelog: one bullet, hand-written into the restored file, guard 32 sections (32 unique) / 64 bullets (64 unique), no repeats. Working tree clean after the chain.

Test-only: one new file plus that bullet. No product code, so shoot and overlap are not required.

Note for whoever merges this and #150 (slice 3): both add a bullet under **Under the hood**, so whichever goes second needs a keep-both re-sync of that one hunk.


Derpimort:
Re-synced onto the RC tip `15a8fc2` (the journeys landed under it). Head is now `1598593`.

**The one conflict, resolved by hand:** my bullet and #152's both sat first under `### Under the hood`. Keep-both, the RC's bullet first, per the standing rule — and verified after the fact rather than assumed: the only diff between this branch's CHANGELOG and the RC's is my own three-line bullet, every RC bullet intact. Guard: 32 sections (32 unique) / 65 bullets (65 unique), no repeats. Never prettier.

**Chain re-run on the merged head `1598593`:** tsc 0 · 180 files, 2847 passed | 1 expected fail | 4 skipped (2852) · coverage 91.64/83.69/91.10/93.70 · lint 0 · format 0 · build 0 · bundle main 394.8/440.0 KB, first load 798.4/1200.0 KB · tree clean. (The one expected fail is #152's `it.fails` pin for #149, now in the RC — not mine.)

CI is green **pinned to this SHA**: I polled until `headRefOid` equalled what I pushed and nothing was pending (6 pending → 4 → clean), per the standing rule. `MERGEABLE / CLEAN`.

For the merger: #150 is separately `MERGEABLE / CLEAN` and needs no re-sync, so the two can go in either order now — whichever is second will need this same keep-both treatment only if the other's bullet lands first.

Derpimort:

Read @1598593 in its own worktree (the re-synced head; RC `15a8fc2` merged in, keep-both on the one Under-the-hood hunk, MERGEABLE).

Test-only: one new file, 283 lines, plus one CHANGELOG bullet. No source touched.

## The control comes first, and that's what makes the rest mean anything

The first pin asserts that the owner's own rolled block **is** a paused thread. Without it, the four empty-rail assertions that follow would pass just as happily against a `paused` list that was broken to always return `[]`. Putting the positive case first is the difference between a pin and a placebo — worth saying out loud, because it's the part most easily left out.

## What I verified myself

Ran on your head: `roll-scaffold-paused` + `week` + `loose-threads-scaffold` + `review-carries-owner-work` → **100 passed**.

My own mutants (sources restored byte-identical after each):

| Mutant | Result |
|---|---|
| W1: the roll gate drops `!placedBy` | **CAUGHT** — 7 of 100 fail |
| W2: a meals-only gate (`placedBy !== 'sustenance'`), so a breather still rolls | **CAUGHT** — 6 fail |
| W3: the rail's `paused` stops requiring `open` | **CAUGHT** — 1 fail, your new case |

W2 is the one I care about: it's the *plausible* wrong fix — someone reading "MEW's seeded meals" as the rule rather than "what MEW placed" — and your mixed-pick case catches it because the breather is in the same list as the lunch. That shape is why this file is worth its 283 lines.

## The finding inside the PR

Your fourth case — a finished follow-up leaving the rail — came from mutating the **rail** rather than the gate, and nothing across 110 loose-threads tests required `paused` to hold open work. A ticked-off follow-up would have sat there as a loose end forever: #123's class, on the surface #147 had just closed. Pinning it here rather than filing an issue is right — today's behaviour is already correct, and the missing thing was a test, not a fix.

That's the second time tonight a mutant found what no reviewer did (mine on #147 was the other). Worth remembering when the next "the tests are green" moment arrives.

## Mechanism, checked in code not prose

`paused` is `blocks.filter(b => b.status === 'open' && rolledTargets.has(b.id))`, and `rolledTargets` is built from `rolledToId` — so the only way scaffolding reaches it is by being rolled. `isRollCandidate` carries `!b.placedBy` and `rollForward` re-checks the gate on the ids it's handed, which your second case exercises by handing it those ids directly. Two independent ends, now both held by tests rather than by reasoning.

Nothing to change. Clean, and the control case is the detail I'd point other people at.

PR #148, opened 2026-09-17
docs(changelog): restore the hand style the prettier reformat overwrote, and write the rule down

Derpimort:
## What
Task `tsk_01m2rbd7wdravzenc3qpenyj4f`, the manager's ruling on my disclosure during the train. Docs-only: `CHANGELOG.md` and `.github/RELEASES.md`, no code.

A `prettier --write CHANGELOG.md` of mine rode into the RC with #140 and reformatted the **whole file**: a blank line after every heading, `*emphasis*` → `_emphasis_`, and one re-wrapped continuation line. Everything renders the same, but it edited the sections of releases that already shipped — and "released sections byte-identical" is the check every verification of these notes rests on (the #105 audit, #112's guard rationale). The root file is deliberately outside the format gate: `pnpm format:check` is `prettier . --check` run from `app/`.

The ruling was to **restore the hand style, not adopt prettier's**, and to write the rule down where the next person will find it.

## How
One deterministic, re-runnable pass (`restore-changelog.py`, run against the RC tip):

- **Released sections** are taken byte-identical from `131a56c`, the last RC commit before the reformat.
- **`[Unreleased]`** keeps every bullet exactly as merged; only the blank line prettier inserted after each `###` heading is removed.

**The base is provably the right one:** the released sections are identical at `131a56c`, at its parent, and at `12ce2b8` (the commit the desktop matrix is green on) — so "as it shipped" and "as it was before the reformat" are the same text.

**Proof this changes nothing but the formatting**, diffing the restored file against the pre-reformat one:

```
restored vs 131a56c: -0  +25
```

Every added line is one of the bullets the train itself merged (#140, #142, #143, #144, #145, #146 and #147, plus #147's own new section heading) — nothing removed, nothing reworded, no released text touched.

`.github/RELEASES.md` gains the rule next to the changelog guard's paragraph: the file is hand-edited, never formatted; prettier rewrites it whole and edits shipped sections; check an edit with `node desktop/scripts/check-changelog.mjs CHANGELOG.md`.

## Calls I made inside the laws
- **The `[Unreleased]` text is untouched.** Restoring "the hand style" could have meant re-wrapping the new bullets too; it doesn't. Every bullet stays exactly as its author wrote and its reviewer read it, so the guard counts don't move.
- **`_emphasis_` is only reverted where it was `*emphasis*` before** — it lives entirely in released sections, which come back wholesale, so no judgement call was needed inside `[Unreleased]`.
[...]

PR #130, opened 2026-09-17
fix(undo): "undo that" reaches MEW's last change in the very next message, and the keyless floor has it too (#120, #118)

Derpimort:
## What
Fixes #120 and #118 (task `tsk_01m2r2vjdben5st1q9gk3wnpkg`). Off RC `dbe3c95`.

**#120.** `speak()` cleared the undo snapshot at the start of **every** message. Only a tap outside any turn (#293: a scenario pick, a drag, the rail place, a confirmed remove, a roll) held it across one. So "undo that", said the message right after a change, answered `nothing to undo yet — I haven't changed the week this turn.` for:
- a keyed or typed change;
- a **picked chip** (a picked chip is its own turn), such as the rescue offer's split;
- a removal picked from a **day chip**.

This is exactly the case `undo_last_action`'s contract names: the user catches a wrong placement and says "no, put it back".

**#118.** The keyless floor had no undo words, so "undo that" fell through to the short-phrase capture and became **an inbox thought** called "undo that", with a when-&-where nudge.

## How
- **`state/store.ts`, one hold for every change.** `snapshotForUndo()` sets `snapshotHolds`, so a snapshot survives exactly the **next** `speak()` entry, whether a turn took it (typed, keyed, a picked chip) or a tap did.
  - A change made in between replaces it, so undo always takes back the **newest** change.
  - A message that changes nothing lets it go: one message only.
  - A second undo still finds nothing.
  - The five tap sites' own `pickSnapshotHolds = true` lines fold into this.
- **`domain/parse.ts`, `UNDO_ASK`.** These are whole-message undo words, read ahead of every capture route, becoming the intent `{ kind: 'undo' }`:
  - `undo`, `undo that`/`it`/`this`, `undo the last change`/`step`/`one`
  - `put it`/`that`/`them back`, `take it`/`that back`, `revert it`/`that`
  - an optional lead-in (`no`, `oops`, `actually`, `wait`, `hmm`) and a closing `please`
  - Near misses stay what they were: `put it back at 3pm`, `undo the laundry`, `take it back to the store`, `revert the deck to 9`.
- **`adapters/model/rules.ts`:** `case 'undo'` calls `exec.undoLast()`, the same executor the keyed model calls. `sanitizeIntent` keeps `undo`.
- **Copy.** The empty-undo line now reads for both paths: `nothing to undo right now — I can take back my last change in your very next message.` The `undo_last_action` description and the `ToolExecutor.undoLast` doc say what undo reaches.

[...]

PR #125, opened 2026-09-17
test(journeys): tonight's features acting on one week together — five real-store journeys, 8 filed gaps pinned

Derpimort:
## What
Task `tsk_01m2r1nfdrpygn5ez1ysz9jg8g`. This PR is **test-only**: one new test file and one *Under the hood* CHANGELOG bullet, on RC `3fe9ebe` (#110 in).

Each of tonight's features was proven on its own, but promotion ships them together. `state/__tests__/rc-journeys.test.ts` drives the **real store** through five journeys, using the keyless floor plus one keyed turn where the tool path differs. At every step it asserts both **the week** and **the chat** the owner sees.

| # | Journey | Features acting together |
|---|---|---|
| 1 | **An out-of-office week, planned into the evening.** A Mon–Wed OOO all-day label, tagged *work* on purpose, sits over Deck, Dinner and a new `block 2h for writing tonight`. | The label stays out of the boot brief (`today: 2 blocks … 1h deep work`). Writing lands **20:00–22:00**: on the hour, after dinner, inside the plannable hours, on an OOO day. `what's on today` shows the label as a label. The 17:20 tick doesn't ask about the rest over dinner. A calendar pull puts `Team sync` on writing, and the rescue offers `split around it` / `roll to tomorrow`. The split, part 2 and load (`work 180, rest 60`) all leave the label out. |
| 2 | **A meeting lands on flexible work.** | The rescue offer lists shift/split/roll; the split is picked (with a breather). `merge my two deck polish blocks` doesn't join across the calendar meeting, and the week stays byte-equal. A keyed turn that moves part 2 and then undoes it restores the week exactly. |
| 3 | **Same-titled lunches across midnight.** | Tuesday 23:50: `remove the lunch at 12:00` shows 4 day chips. After midnight, tapping **tomorrow 12:00** is refused (`That choice was offered on Tuesday …`) and nothing is removed. Asked again on Wednesday, `tomorrow 12:00` removes **only Thursday's** lunch. |
| 4 | **Friday: a split pair, the room offer, rolling forward.** | A split around a call. `block inbox sweep, block errands`: the room offer names `(inbox sweep, errands)` and pads exactly those two (60→70). Rolling the draft forward, then one split part, moves only those blocks; Friday's other rows stay byte-equal, and next week holds the copies. |
[...]

Issue #124, opened 2026-09-17
[dev] The remove ask for 3+ same-titled blocks says "say both to drop them all" (the chip already says "all of them")


Derpimort:
## What happens
The remove ask for 3 or more same-titled blocks says **"both"**.

**Repro (real store, keyless floor):** Tuesday, with Lunch 12:00–12:45 on Tuesday, Wednesday and Thursday.
- `remove the lunch at 12:00` → `3 "lunch" blocks ahead — the 12:00 (12:00–12:45), the 12:00 (Wednesday 12:00–12:45), or the 12:00 (Thursday 12:00–12:45)? Tell me which, or say "both" to drop them all.`
- Its chips are `today 12:00 · tomorrow 12:00 · thursday 12:00 · all of them`.

The prose is fixed at "both" (store.ts ~4800) whatever the count, while the chip already says "all of them" once there are three or more.

## Expected
With 3 or more candidates the line reads `… or say "all of them" to drop them all.`, matching the chip. With exactly two, "both" stays.

Found by the RC cross-feature journeys (`rc-journeys.test.ts`, test-only). It's pinned there with `it.fails`.

PR #152, opened 2026-09-17
test(journeys): round 2 — the features that landed after #125, acting together

### LINKED ISSUE #149: [dev] Undo after a retag says the blocks were put "back where they were" — a retag moves nothing

Derpimort:
## What
Task `tsk_01m2rdpd9a4pj4mk1nbp4trfha`. Round 1 (#125) found **eight** real bugs by running tonight's features together on one week. Since then the RC gained batch shift and retag, split and merge-back, undo reaching the previous message, the remove ask's typed answers, one length on a re-plan, and MEW's own scaffolding kept out of carried work, open items and the rail. Nobody had run **those** together.

Test-only: a new `rc-journeys-2.test.ts` (its own harness, so no lane shares a file) plus one CHANGELOG bullet. **One bug found, filed as #149, not fixed here.**

## The journeys
Each drives the real store on the keyless floor and asserts **the week and the chat** at every step.

**6 · a split pair, MEW's own scaffolding and a protected rest meet a batch shift.** Sustenance on, so MEW's lunch, dinner and a pacing breather share the day with the owner's Deck, a calendar call inside it and a protected evening walk.
- the split works around the calendar event and moves **MEW's own breather** aside, not the owner's work;
- the batch offers first and changes nothing, listing the one block that can move and naming both that can't — the calendar event, and the half of the split whose new time *would sit over* it;
- the pick moves exactly what it listed, and the protected rest it now runs over is named in #122's words, never "flexible" (#141);
- one "undo that" puts the whole day back, scaffolding untouched throughout.

**7 · three same-titled lunches, answered in words.** The ask names the all-chip in #124's words; a count that doesn't fit (`all 2`) changes nothing and brings the ask's chips back (#140); `the thursday one` removes exactly one through the chip's own path, so the log shows what MEW acted on; `undo that` returns it (#130). Nothing lands in the inbox at any step.

**8 · a retag over a day holding what MEW may not touch.** Four "calls" match; the calendar event and the repeating occurrence are **named, not silently dropped**; the pick changes only tags, no block moves a minute; one undo puts both tags back.

**9 · Friday evening, on a week MEW helped build.** The weekly review carries only the owner's work (#138); rolling one moves only it and says so; and once the seeded dinner's window has passed, neither the day's open items (#144) nor the loose-threads rail (#147) holds it, while the owner's own "lunch with sam" is in both. First time those three exclusions have met on one week.

A retag moves nothing, so the receipt describes something that never happened. The **week** is right (both tags return, times untouched, the calendar event and the repeating occurrence never touched) — only the sentence is wrong. `execUndo`'s summary describes every *changed* block with the move vocabulary, and since #142 a changed block can differ only in its tag.

Filed as **#149** with the repro and a suggested shape; pinned here as `it.fails` in journey 8, step 4, so it goes green the moment the fix lands. No product code in this PR, per the task.

## Calls I made inside the laws
- **The failing pin asserts the absence of the wrong words**, not a guess at the right ones — whoever fixes #149 chooses the wording, and this pin won't fight them.
- **Journey 9 reads the domain helpers directly** (`openItems`, `looseThreads`, `dayClear`) rather than rendering the rail: the rule lives there, `ThreadRail` renders it, and a jsdom render would pin React, not behaviour.
- **Journey 6 keeps sustenance on** so MEW's own blocks are real, not hand-made fixtures — the scaffolding has to be what the scaffold actually places for the journey to mean anything.

## Gates
Test-only, so shoot and overlap aren't required (no `ui/` or `scripts/` change). The one expected fail is #149's pin.

```
head 507690a (off the RC tip b2ed943, merged in)
install exit 0
tsc exit 0
lint exit 0
format exit 0
coverage exit 0
build exit 0
bundle exit 0
tree: 0 change(s)
 Test Files  179 passed | 2 skipped (181)
      Tests  2837 passed | 1 expected fail | 9 skipped (2847)
Statements   : 91.64% ( 8417/9184 )
Branches     : 83.69% ( 6780/8101 )
Functions    : 91.1% ( 1730/1899 )
Lines        : 93.77% ( 7156/7631 )
| ✅ | **main** (entry + its static non-vendor chunks) | | **394.8 KB** | **440.0 KB** |
| ✅ | **first load** (eager) | | **798.4 KB** | **1200.0 KB** |
check-bundle-size: OK
```

CHANGELOG: one bullet under "Under the hood", hand-written into the restored file (#148). Guard: 32 sections (32 unique) · 64 bullets (64 unique), no repeats.


Derpimort:
## Peer review (coderpb): `507690a`

**Verdict: APPROVE. Zero 🟡.** Test-only, and it does the thing round 1 proved is worth doing — it found a real bug (#149) and pinned it without touching product code.

Every journey asserts the **week and the transcript** at each step, with exact strings rather than `toMatch` on a fragment, so a reply that drifts a word fails here. That's the right standard for a journeys suite.

### The central claim, tested rather than accepted

Journey 9 says it runs #138, #144 and #147 together for the first time. I checked that by reverting each fix in turn (source restored byte-for-byte after each):

| reverted | result |
|---|---|
| #147 — the rail's `slipped` filter | **1 test fails** |
| #144 — `openItems` | **1 test fails** |
| #123's roll gate — the review's carried list | **1 test fails** |

So the journey is genuinely load-bearing on all three, not just arranged next to them.

### The #149 tripwire works, and journey 6 guards the fix

I simulated the fix — changed the `changed` branch of the undo summary so it no longer says "back where they were" — and re-ran:

- the `it.fails` pin **flips to failing**, which is exactly what should happen: whoever fixes #149 must convert it to `it`. The tripwire is real.
- **journey 6 also failed**, and that is a *feature* of this file, not a hazard. Journey 6 pins the undo receipt for a genuine **move** (`put … back where they were`). My simulated fix was crude — it changed the wording for every changed block — and journey 6 caught the over-reach immediately. So when you fix #149: the shape in your issue (split `changed` by what actually differs) keeps journey 6 green, and a fix that changes the *move* wording too will not get past it. That's a useful constraint to have written down before the fix, and it's one more reason not to hand-wave the receipt.

### On `it.fails`, said plainly

An `it.fails` body passes when it throws **for any reason**, so in principle a broken fixture in the pin's setup would keep it green while proving nothing. Here it is mitigated by construction: the pin's setup (`fresh` → retag ask → `pick('do it')` → `undo that`) is the same flow the sibling test in the same describe asserts step by step with exact transcripts, so a setup break fails loudly there before it can hide here. I'd keep the pin's body minimal exactly as it is — adding more assertions would give it more ways to throw for the wrong reason.

### Cross-PR check, since three PRs are open at once

Journey 8 retags a selection containing a **repeating** block, which is precisely what my #150 (slice 3) changes. I copied your file into my slice-3 worktree and ran it against that code: **4 passed | 1 expected fail — unchanged.** The reason it's safe: `c4` is the only occurrence of `r1`, so `seriesMembership` counts 1, and slice 3 deliberately asks nothing when the three answers would collapse into one. Your transcript stays right either way.

Worth knowing for later: once #150 lands, a journey that retags a selection holding a **multi-occurrence** series will get the scope question first, and its transcript will differ. Slice 3's own suite covers that shape, so nothing is missing here — I'm just flagging it so a future round-3 journey doesn't get written against the old flow.

### The one merge hazard

Your bullet and my #151's bullet insert at the **same line** — both are the first bullet under `### Under the hood` (line 262). Whichever merges second gets a conflict on that one hunk; the resolution is keep-both, RC's bullet first, hand-edited, then re-run `check-changelog.mjs`. Nothing else in the three open PRs overlaps: #150 is `app/src` + CHANGELOG, #151 is one new test file + CHANGELOG, yours is one new test file + CHANGELOG.

### Gates

I ran the suite myself: `4 passed | 1 expected fail (5)`, matching your numbers. CI on this head: typecheck, lint, bundle, release-guard green (unit and lint-full were still running when I looked — I'll confirm they finished before merging).

Also: thank you for the TS6133 note. It is the same trap in the opposite direction from my `Standup` fixture — a copied harness brings helpers the new file never uses, and the gate only tells you after the coverage step's minutes. I pruned mine to what it uses in the round-3 pins for exactly that reason.

Issue #139, opened 2026-09-17
[dev] Typing a live chip's own label does nothing outside remove asks: "roll to tomorrow" and "tomorrow 9:15" become inbox thoughts, "do it" and "just this one" get the help line


Derpimort:
## What happens
While chips are live, **typing a chip's own label** does nothing on the keyless floor, for every chip family except the remove ask (#131 fixes that one). Only tapping works, and some typed labels land in the inbox as thoughts.

**Audit (real store, keyless floor, RC `333ecfc` plus #131):**

| Family | Typed label, chips live | Result |
|---|---|---|
| Rescue offer (#286) | `roll to tomorrow` | `Captured "roll to tomorrow".`: an inbox thought, nothing moved |
| #116 no-room offer | `tomorrow 9:15` | `Captured "tomorrow 9:15".`: an inbox thought, nothing placed |
| Batch confirm (#110) | `do it` | the generic `I can place that for you — try …` line, nothing moved |
| Split series-scope ask (#343) | `just this one` | the generic help line, nothing split |

Families not yet audited, likely the same: drift choices (`keep both`, `drop Groceries`), the room offer (`give them room`, `leave as-is`), which-block chips for move/split (`the 8:30`), and the day-load/trim offers.

## Expected
A typed reply that **is** a live chip's label acts exactly like tapping that chip: the same `pickChoice` path, with #94's pick-time re-check and the chips retired. It never becomes an inbox thought or a help line. #131's resolver (`typedRemoveAnswer`, whole-message, only while the chips are live, only when the words point at exactly one chip) is the shape to generalise, with each family's exact labels first. Remove asks also keep their extra words (`both`, `all of them`, `the thursday one`).

Found auditing the other chip families for #131, as its brief asked.

Derpimort:
## #139 slice 2 audit — coderpa's half: the drift choices and the room offer

Probed on the real store at the RC tip `958732e` (post-#153), each label **typed** at the live ask, never tapped. Probe files removed afterwards; tree clean.

**Result: both families are fully covered. Nothing to fix here — the finding is that there is nothing to find, which is worth a pin rather than a blank.**

### Drift choices (`offerDriftChoices`)
Setup: `block 1h for the release review today at 14` over a flexible Groceries 14:00–15:30, walls either side. Chips offered: `move release review to tomorrow 9:00` · `drop Groceries` · `keep both`.

| typed | what MEW did | inbox | week |
|---|---|---|---|
[...]

Issue #131, opened 2026-09-17
[dev] The remove ask says to "say both" / "say all of them", but typing it does nothing: the words become inbox thoughts or the generic help line


Derpimort:
## What happens
The remove ask tells the owner to **say** a word, but on the keyless floor saying it does nothing useful. Only the chip works.

**Repro (real store, keyless floor, RC `d7f57d9`):** Tuesday, with Lunch 12:00–12:45 on Tuesday, Wednesday and Thursday.
- `remove the lunch at 12:00` → `3 "lunch" blocks ahead — … Tell me which, or say "both" to drop them all.` (#129 changes the word to "all of them").
- Then, typed:
  - `all of them` → `I can place that for you — try "block thursday morning for the deck", …`. Nothing is removed.
  - `both` → `Captured "both".`, an **inbox thought called "both"**.
  - `all` → `Captured "all".`
  - `remove all of them` → `Captured "remove".`, an inbox thought called **"remove"**.

The all-chip itself (`remove all lunch`) works. The typed words have no context: the floor doesn't know a remove ask is standing.

## Expected
Either:
- the words the ask tells the owner to say act like its chip, the way a typed reply to a standing ask resolves against that ask; or
- the line doesn't send the owner to typing ("… or pick "all of them" to drop them all").

Either way, a bare `both` / `all` / `remove all of them` never becomes an inbox thought.

Found while peer-reviewing #129 (#124). Pre-existing on the RC.

Issue #116, opened 2026-09-17
[dev] A keyless plan that cannot fit before the plannable end lands in the past: at 20:46, "block 2h for writing" → today 8:00–10:00


Derpimort:
## What happens
A keyless plan that doesn't fit between **now** and the end of the plannable hours lands **earlier the same day**, in time that has already passed.

**Repro (real store, keyless floor):** Tuesday 20:46, plannable hours at their default 8:00–22:30, the morning empty.
- `block 2h for writing` → `Done — today 8:00–10:00 is held for writing. That's your 1st hour-plus work block this week. Tucked a 15-min breather into today at 10:00.`
  - The block sits at 8:00–10:00 **today**, twelve hours in the past, with a breather at 10:00 after it.
- The same happens with `block 2h for writing tonight`.
- A plan that fits, such as `block 1h for email` at 20:46, lands correctly at `today 21:00–22:00`.

## Expected
A placement never starts in the past. If today can't hold it before the plannable end, MEW says so plainly and names the open air past those hours and the nearest options, as `find_slot` and the keyed plan already do ("No 60-min window today fits inside the hours I plan in (8:00–22:30). Open air today: …"). Or it offers tomorrow. It never lands silently in the morning that has already gone.

## Where (from reading the code)
- In `execPlan`, `scoreSlots(…).find((c) => c.dayKey === key)` finds nothing today once the day is too short.
- `week.place` then runs first-fit from the plannable **start** (8:00) with no floor at now (store.ts ~2808–2890, week.ts ~388).

Found by the RC cross-feature journeys (`rc-journeys.test.ts`, test-only). The journey pins it with `it.fails` and this issue number.

PR #110, opened 2026-09-17
feat(batch): change several blocks at once — a wide change is offered with the exact list first, one undo (#75 slice 1)

Derpimort:
Refs #75 (**slice 1**; the issue stays open for the rest). Task `tsk_01m2qytgm7y41yc72h88mp8vet`. Head `27395cb` on RC `d256b88`. Commits:
- `de311a1`: the slice
- `f6e4823`: live-probe fixes
- `1be30da` + `4d4a8c8`: coderpa's review round 1, [reply here](https://github.com/Derpimort/mew/pull/110#issuecomment-5717482205)
- `5a61d94`: review round 2, [reply here](https://github.com/Derpimort/mew/pull/110#issuecomment-5717736312)
- RC merges

## What the owner can say now
- **"push everything after 3pm back an hour"** · "pull all work before noon 30 min earlier"
- **"move all of today's work to tomorrow"** · "move all thursday's deck review blocks to friday"
- Keyed models get the same thing as `batch_blocks`.

Before this, the model called `move_task` once per block, with no consent step for a wide change and one undo snapshot per call.

## The law this is built around: a wide change is offered first, with the exact list
```
you ❯ push everything after 3pm back an hour
mew ❯ move 3 blocks 60 min later? Deck 15:00→16:00 · Email 17:00→18:00 · Notes 18:00→19:00. Client call 15:30 (fixed) and Quarterly planning 20:00 (from your calendar) stay where they are.
      [do it]  [not now]
```
Nothing changes before the pick. **do it** re-asks in the keyless grammar's own words with the count and a **list token**: `push everything after 15:00 today later by 60 min — yes, all 3 · k7f2`.
- The rules floor applies exactly that.
- A keyed model reads it and calls again with `confirmCount: 3, confirmToken: 'k7f2'`.
- **If the plan no longer moves exactly that list**, MEW offers again: *"the week changed since then — move 3 blocks …?"*. The trigger can be the week changing, a drag or inbox tap in between, or the other floor reading the selector differently. Nothing moves.

## The change
- **`domain/batch.ts` (new, pure):**
  - **`selectBatch`** picks what `list_blocks` shows that day, all-day labels aside, narrowed by the start window, tag and title words.
  - **`planBatch`** decides what moves where and what stays put, and why:
    - `calendar`, `fixed`, `done` and `repeating` blocks never move (series edits ask their scope first, per #343);
    - `off-day`: a shift that would run past midnight stays put;
[...]

PR #153, opened 2026-09-17
feat(chips): typing a live chip's own label does what tapping it does — the four audited families (#139 slice 1)

Derpimort:
Only a remove ask read typed answers (#131). Everywhere else, typing what was **on screen** did nothing: `roll to tomorrow` and `tomorrow 9:15` became thoughts in the inbox, `do it` and `just this one` got the generic help line. All four audited families now answer to their own labels — and so does every family that ships chips, because this is one floor under all of them rather than four special cases.

## What it does

`typedChipLabel` is pure: it says which chip on which message. The store routes that through **`pickChoice`**, so the tap's own pick-time re-checks (#94's day re-check, the drift-drop exactness guard) and retiring the chips are not re-implemented anywhere — a typed answer and a tap are literally the same path from that point on.

## Deliberately narrow, per the brief's "each family's exact labels first"

- the **whole message**, exactly, with case, surrounding space and one trailing `.`/`!`/`?` aside. `do it after the gym instead` is conversation, not a tap.
- only while the chips are **live**, and only the newest ask.
- only when the words point at exactly **one** chip — an ambiguous answer is never resolved by chip order.
- **after** the remove ask's own richer reader, so `both` for three blocks still gets its plain answer rather than becoming a pick.

### One ordering detail that is load-bearing

The hook sits **before** the typed message is posted. #254's law is that a newer user turn retires the chips, so asked after the post, the owner's own words would supersede the chips they name and no label could ever match. That is pinned in the pure tests with the reason written beside it.

## Tests: 15, driven through the real offers

The four families are produced by the real code paths, not hand-built fixtures: a pulled meeting (`simulatePull`), a day full to the end of the plannable hours, a three-block sweep, and a split of a repeating block.

| family | typed | result |
|---|---|---|
| rescue offer (#286) | `roll to tomorrow` | the block rolls; nothing captured; chips retired |
| no-room offer (#116) | `tomorrow 8:00` (the label it offered) | placed there tomorrow |
| batch confirm (#110) | `do it` | `Moved 3 blocks 30 min later today — …` |
| series-scope ask (#343) | `just this one` | that occurrence splits; tomorrow's untouched |

Plus the laws: a label typed once the chips are spent comes back as an **ordinary turn** (not swallowed), a label buried in a longer sentence is not a pick, a remove ask keeps its richer reading, and #94 still answers `That choice was offered on Tuesday …` for a label typed after midnight.

## Mutants

Each run alone against the three suites, sources restored byte-for-byte:

| mutant | tests failing |
|---|---|
| the feature removed (the RC's own behaviour) | 5 |
| a spent chip still answers to its label | 2 |
| an ambiguous label takes the first match | 1 |
| a label found anywhere in the message counts | 2 |
| the match becomes case-sensitive | 1 |

**One mutant is equivalent and I am declaring it rather than hiding it:** running the generic reader *ahead* of the remove ask's changes nothing observable, because the only inputs the remove reader answers differently (`both` for three) are not chip labels, so they fall through to the same clarify. The order still stands as written — it is the honest expression of "the remove ask's reading wins" — but no test can distinguish it today.

### Two things my own tests caught, recorded because they are the kind of thing that makes a green suite meaningless

- The copied harness's `lastMew()` returns the **message**, not its body, so `expect(lastMew()).not.toContain('Captured')` was passing for the wrong reason — a false green in the very assertion that proves the inbox bug is fixed. There is now a `lastBody()` helper and a comment saying why.
- My first pass proved the feature worked but **not one of its limits**: four of six mutants survived. The spent-chip, ambiguity, whole-message and case guards each have a test now because of that run, not because I thought of them while writing the code.

Also: the copied harness brought five unused helpers with it (`TS6133`) — coderpa's slip from an hour ago, caught the same way. Pruned.

## Gates — RC `c5beb34`

```
tsc exit=0
coverage exit=0      Test Files  181 passed | 2 skipped (183)
                           Tests  2854 passed | 1 expected fail | 9 skipped (2864)
                     All files | 91.66 | 83.76 | 91.11 | 93.70
                     main 395.2 KB / 440.0 KB · first load 798.8 KB / 1200.0 KB
```

The one expected fail is #152's pin for issue #149, already in the RC. CHANGELOG: one new section, hand-written, guard 33 sections (33 unique) / 66 bullets (66 unique), no repeats. Tree clean.

Shoot and overlap are not required: `app/src` and `CHANGELOG.md` only, no `ui/`, no `scripts/`. The chips themselves are unchanged — this adds a second way to answer them, and the existing overlap proof covers how they render.

## Not in this slice

The unaudited families (drift choices, the room offer, which-block chips, day-load/trim) already work through this same resolver, since it reads whatever labels the live ask carries — but I have only *tested* the four the issue audited. The per-family extras beyond exact labels (`keep both` phrased as "keep them both", a time typed as "9.15") are the next slice, and the issue's own shape suggests them.


Derpimort:
## Re-synced onto `69a6a17`, and the manager's question answered — head `cc40f1d`

**Does the resolver cover #150's batch scope ask, or only #343's split scope ask?**

**It covers it, and there is now a test that says so.** Answered by running it, not by reading the code:

```
push all health after 4pm today later by 30 min   → Gym repeats — which do you mean?
                                                     [just this one] [this & the ones after] [the whole series]
the whole series                                  → Moved 2 blocks 30 min later —
                                                     Gym today 18:00→18:30 · Gym tomorrow 18:00→18:30.
```

The reason it works is the shape of the resolver rather than luck: it reads **whatever labels the live ask carries** and never keys on a family. That is why a fifth family could land after this branch forked and still be covered — and why the unaudited families (drift choices, the room offer, which-block chips, day-load/trim) are covered too, though I still only *claim* what I have tested.

The pin asserts the **whole** receipt sentence, so it holds #150's copy fix from the other side as well: each row named with its own day, no "today" claim over a change that also moved tomorrow's.

**Pulled into slice 1**, per your preference, and I agree with the reasoning: the owner will meet the batch scope ask first — it is the newest thing in the RC — and a fix that taught "typing a chip label works" and then failed on that ask would be worse than shipping nothing.

## One thing I did wrong twice, and am saying so twice

My inserted test landed **outside** the describe again — the same slip I disclosed on the batch suite an hour ago, now in a second file. A top-level `it()` runs, so nothing complained. It is inside the describe now. Three doc comments that my earlier harness pruning had orphaned (they described helpers I removed) are gone in the same pass.

The pattern is clear enough to name: when I insert a test by anchoring on `describe(`, I get file-level placement unless I anchor on the describe's **closing** brace. I am now checking `grep -c "^it("` is zero before I push a test file.

## Gates on the re-synced head `cc40f1d`

```
tsc exit=0
coverage exit=0      Test Files  183 passed | 2 skipped (185)
                           Tests  2889 passed | 1 expected fail | 4 skipped (2894)
                     All files | 91.69 | 83.83 | 91.16 | 93.78
                     main 396.5 KB / 440.0 KB · first load 800.0 KB / 1200.0 KB
```

The one expected fail is #149's pin, which #155 takes to zero. CHANGELOG re-checked after the merge: guard 33 sections (33 unique) / 67 bullets (67 unique), no repeats, and the only difference against the RC is my own section. Tree clean.

@coderpa — ready for you whenever #149 is out of your hands. Per the manager's order I am reviewing #155 first, and #153 merges second.

Derpimort:

Read @cc40f1d (re-synced onto `69a6a17`, so the tree contains the fifth chip family this question is about).

## The manager's question, answered by test — and the answer is structural, not lucky

Does the resolver reach **#150's batch scope ask**, or only #343's split one? They speak two of the same words and are different asks.

`typedChipLabel` reads the labels the **live ask carries** and never looks at a family, a `nudgeType` or a reply shape. So the coverage isn't a list that has to be kept in step with the product — a family that lands *after* this branch forked is covered by construction, which is exactly what happened here. Your pin proves it on the real ask and asserts the **whole receipt**, so it holds #150's copy fix from the other side too.

That's the difference between "we remembered to add it" and "it could not have been missed", and it's why the answer is worth more than the pin.

## Verified here, baseline first

Baseline on your head: `typed-chip-labels` + `choices` + `typed-remove-answer` + `batch-series-scope` → **40 passed**. (Baseline before mutating, per tonight's new rule — which exists because I skipped it once.)

My own mutants, sources restored byte-identical after each:

| Mutant | Result |
|---|---|
| Z1: a spent or superseded ask still answers | **CAUGHT** (2) |
| Z2: an ambiguous label picks the first chip | **CAUGHT** |
| Z4: the label is read **after** the message is posted | **CAUGHT** (6) — the ordering claim, held hard |
| Z5: case and trailing punctuation must match exactly | **CAUGHT** |
| Z3: the generic reader runs **before** the remove ask's | **SURVIVED** — your declared equivalent |

**On Z3 I checked the equivalence rather than accepting it**, since "declared equivalent" is where a real gap can hide. The two orders agree on every input either can answer:
- a 2-lunch ask, typed `both` → label-exact under both orders, same chip;
- a 3-lunch ask, typed `both` → no chip reads "both", so the generic reader returns null and the remove reader gives its count answer either way;
- `the thursday one` → no exact label, same fall-through;
- a label typed exactly (`all of them`) → same chip from both readers.

Where one fires the other returns null, so nothing observable moves. The claim holds.

## Shape

Routing through `pickChoice` rather than re-implementing the tap is the right call: #94's pick-time re-check, the drop-chip guard and chip retirement stay in one place, and the log shows the chip's own reply, so the owner sees how their words were read. Running **after** `typedRemoveAnswer` keeps the remove ask's richer readings (a count that doesn't fit still gets its plain answer). Narrow on purpose — whole message, exact, one match — so an ambiguous answer is never resolved by chip order.

And the ordering detail is the load-bearing one: reading the label **before** the message is posted, because #254 retires chips on a newer user turn. My Z4 mutant takes six tests down, which is the right amount of noise for a line that would otherwise look like a style choice.

## The second top-level `it()`

Second occurrence, second disclosure, and this time with the pattern named — anchoring an insert on `describe(` lands the test at file level; anchor on the closing brace, and check `grep -c '^it('` is zero before pushing. Naming the *cause* rather than fixing the instance is what stops a third.

Nothing to change. Merging it third, after #155 and #154, per the order.

PR #154, opened 2026-09-17
test(batch): pin the half of spansDays that reads the blocks which stay (#150 review), and put two orphaned tests back in their describe

Derpimort:
coderpa's optional 🟢 from the #150 re-review, closed by the author of the gap rather than carried into journeys round 3. It was a real hole in the net around my own fix: `spansDays` deliberately counts the **skipped** blocks' days as well as the moves', and a mutant reading only `plan.moves` survived the entire suite.

## The shape that needs it

coderpa's, exactly as described. A Tue/Wed/Thu Gym series answered **"the whole series"**, with a calendar event at 18:30 on **both** Wednesday and Thursday:

- those two occurrences skip as `lands-on` (a calendar event is never moved), so **only Tuesday's moves**
- the **moves** are therefore single-day, while the sentence still crosses days *because of what stayed*

Today that reads:

```
move 1 block 30 min later? Gym today 18:00→18:30.
Gym tomorrow 18:00 (would sit over Board call 18:30–19:00) and Gym Thursday 18:00
(would sit over Board call 18:30–19:00) stay where they are.
```

With the moves-only mutant it goes back to two **identical** rows under a claim that only covers today — the owner cannot tell which occurrence stayed for which reason, and cannot check the receipt against their week afterwards. That is the same defect the 🟡 was about, on the half nothing held.

The pin asserts the offer's exact wording and then the week after the pick (one occurrence moved 30 minutes, two kept their place, the calendar events untouched).

**Mutant:** `spansDays` reading only `plan.moves` → **1 test fails** (this one), where before it survived all 10 in the file. `store.ts` restored byte-identical after the run.

## Disclosed, not quietly fixed: two tests were outside their describe

While working in the file I found that two tests sat at **file top level** rather than inside `describe('#75 slice 3 — a sweep over a repeating block asks first')` — and one of them is mine from #150, already merged. A top-level `it()` still runs, which is why nothing complained, but it reports with no suite name, so a failure reads without the context that tells you what broke.

Both are inside the describe now, indentation and all. **No test text changed** — the diff is indentation plus the two moves, which is why it reads as 63 insertions / 25 deletions for what is really one new test. The duplicated `fixtures` banner my line surgery left behind is cleaned up in the same pass.

I would rather say this than let it look like the file was always tidy: I introduced it, the gate did not catch it, and reviewing my own diff is what found it.

## Gates — RC `69a6a17`

```
tsc exit=0
coverage exit=0      Test Files  182 passed | 2 skipped (184)
                           Tests  2872 passed | 1 expected fail | 9 skipped (2882)
                     All files | 91.67 | 83.76 | 91.14 | 93.70
                     main 396.0 KB / 440.0 KB · first load 799.6 KB / 1200.0 KB
```

The one expected fail is coderpa's #149 pin, already in the RC, and it goes to zero with their fix. CHANGELOG: one bullet, hand-written, guard 32 sections (32 unique) / 67 bullets (67 unique), no repeats. Tree clean.

Test-only — one test file and that bullet, no source touched — so shoot and overlap are not required.


Derpimort:
**Peer review (coderpa) — APPROVE. 0 🟡, 0 🟢. The mutant is dead, and I killed it myself.**

Read @3941e16 (off `69a6a17`). Test-only: one new test, one CHANGELOG bullet, no source.

## The independent check the ruling asked for

The manager's condition was that the reviewer who found the surviving mutant proves the pin kills it — not the author. So I ran it here, on your head:

```
baseline                                   Tests  10 passed (10)
spansDays = moves-only (my mutant)         Tests  1 failed | 9 passed (10)
  × the days that cross can be the ones that STAY: skipped rows carry their day too
store.ts restored byte-identical
```

**Before your pin that mutant survived all ten tests in this file. Now it fails exactly one — the one written for it, and nothing else.** That's the shape a pin should have: precise, not incidental.

## The pin itself

Your fixture is the reachable case and it's built the honest way — two *calendar* events at 18:30 on Wednesday and Thursday, so both occurrences skip as `lands-on` and only Tuesday's moves. The moves are single-day; the sentence crosses days only because of what **stayed**. That's exactly the half a moves-only reading gets wrong.

Three assertions, all earning their place: the moving row names its day; the two stays are told apart by theirs, as one exact string; and `not.toContain('30 min later today?')` closes the false claim directly rather than by implication.

## The disclosure

Two tests were sitting at file top level, outside their `describe` — one of them yours, already merged in the RC. You found it by reading your own diff, said so before anyone asked, and moved both inside without touching a word of their text. I verified that claim rather than taking it: the test titles in this file differ from the RC's by **exactly one line** — your new test — and there are now **no** top-level `it(` calls. So the 63/25 line count really is one new test plus indentation.

A top-level `it()` runs, which is why no gate complained; it just reports without the suite name that tells you what broke. Worth fixing, worth disclosing, and exactly the right size of noise to make in a PR.

## Merge note

Approved, and I'm **holding the merge until #149 lands** — not because of anything here. Both PRs add a CHANGELOG bullet, and #149 is the one that takes the RC's single expected fail to zero; landing this first would make that PR pay a keep-both re-sync for no gain. The moment #155 is in, this goes next under my own chain.

Issue #73, opened 2026-09-17
[dev] Split a block around a time or a meeting: the split op from #16's calendar command surface


Derpimort:
## What & why

Split from #16 (calendar command surface, the standing "more every RC" track). v0.6 shipped **resize**, **duplicate** and **relative move** (`resize_block`, `duplicate_block`, `move_relative`; Derpimort/mew-archive#353). **Split** is one of the three remaining ops.

The owner wants to say *"split the deck around the 1pm call"* and have one block become two, with the time in between left free. The logic half-exists already, but only inside the rescue flow:
- `domain/rescue.ts:188` `splitReply` / `:202` `parseSplitAsk` read a rescue chip's exact reply (`split the deck around 13:00-13:45, keep 45m after`).
- `adapters/model/rules.ts:284` routes that one phrasing.

There is no model tool, and no keyless route for a naturally worded split.

## Architecture placement
- `adapters/model/tools.ts` → a `split_block` tool: `query`, `at` (disambiguates same-named blocks), and **either** `aroundStartMin`/`aroundEndMin` (a gap) **or** `aroundQuery` (the block to split around, e.g. the 1pm call).
- `state/store.ts` → `execSplit`, through the one mutation path (`setBlocks` + `snapshotForUndo`). It reuses the rescue split's geometry, so the rescue chip and the tool share **one** implementation.
- `domain/parse.ts` + `adapters/model/rules.ts` → a keyless route for "split X around <time>-<time>" and "split X around <block>".
- `domain/week.ts` → any pure helper the two paths share.

## Product laws
- Tools-only; one mutation path; undoable.
- **External events are never split.** A `[calendar]` block isn't MEW's to cut. The block split *around* may be external, and that's the common case.
- Fixed-time blocks are scheduled around, never over: the second half lands only in time that's genuinely free.
- Recurrence: splitting one occurrence of a series asks this / following / series first (#343's scope chips), never silently.
- Positive voice.

## Acceptance criteria
- [ ] "split the deck around the 1pm call" (keyless **and** keyed) leaves two blocks with the call's span free between them, with the total length kept; pinned through the real store
- [ ] The rescue chip's existing reply still splits byte-identically (one shared implementation, pinned)
- [ ] A `[calendar]` block is never split; asking says so plainly; pinned
- [ ] A series occurrence asks scope first; pinned
- [ ] tsc · coverage · lint · format · build green; CHANGELOG line

## Out of scope
Merge and batch (sibling issues). Splitting across days.

[...]

PR #155, opened 2026-09-17
fix(undo): the receipt names what actually came back (#149)

### LINKED ISSUE #149: [dev] Undo after a retag says the blocks were put "back where they were" — a retag moves nothing

Derpimort:
## What
Fixes #149 (task `tsk_01m2re8ejqgmqq3r0f57hkt366`). Off the tip `69a6a17`.

`execUndo` described **every** changed block with one sentence — "put X back where it was" — so only a real move was told truthfully. The filed bug was a retag; coderpb's probe on the real store found its reach, and my own gates found one more:

| the change | before | now |
|---|---|---|
| a **move** | put Deck back where it was | *unchanged* |
| a **resize** | put Deck back where it was | put Deck back to its **old length** |
| a **rename** | put Deck back where it was | **called it Deck again** |
| a **retag** (as filed) | put Deck back where it was | put Deck **back to work** |
| a **split's own undo** | …put Deck polish back where it was | …put Deck polish back to its **old length** |

In every case the week was restored correctly. Only the sentence lied — and undo is the safety net under everything else that shipped tonight, so a net that misreports what it caught teaches the owner not to trust it.

## How
`state/store.ts`: the summary reads the **fields** rather than the object identity that `changed` is built from, and names the difference it finds — where it sits, its length, its name, or its tag.

- **Precedence, pinned rather than left to reading order:** time, then length, then name, then tag. A block that moved *and* was retagged takes the larger fact — where it sits.
- **Anything else that differs** falls back to "back **as it was**", which is true of every field, so a future field can't quietly inherit the move wording.
- **A genuine move is byte-identical**, which journey 6 and the split pins both hold.

## The one I didn't expect: the split
My chain failed on two existing pins in `undo-next-message.test.ts`, which expected `put Deck polish back where it was` after undoing a split. A split **shortens** the block it splits — its start never moves — so those pins were encoding #149's own class on a path the issue never named.

I changed both expectations rather than special-casing the split, and left a note at the top of that file saying why: **an existing file's changed expectation should never be quiet.** The receipt's first clause — "took back the two blocks I'd just placed", which is what the split actually did — is untouched.

## Tests
New `state/__tests__/undo-receipt.test.ts` (7, real store, keyless). The artefact under test is a sentence the owner acts on, so **every pin asserts the whole sentence**:
1. a move reads exactly as before — the control, and the wording the others were borrowing;
2. a resize says the length came back;
3. a rename says it in the name that came back (not the one being undone);
4. a retag names the tag it came back to — the case as filed;
5. two blocks retagged → `Undone — put two tags back.`;
6. a block that both moved and was retagged takes the move wording;
7. **a batch move to another day**, which keeps every block's clock by design, so only the day differs — added after coderpb's review found that dropping the `dayKey` half of the move test survived all 25 tests. A keyless single move *re-times* the block, so `startMin` differs anyway and the day half was never exercised. With their mutant applied, this pin is the only failure.

`rc-journeys-2.test.ts`: journey 8's `it.fails` pin for #149 is now an **ordinary test** asserting the whole sentence. **The RC's expected-fail count goes to zero.**

**Mutants (all caught; `store.ts` restored byte-identical after each):**

| Mutant | Fails |
|---|---|
| every changed block called a move again — the bug | 6 |
| the tag read before the move, so the larger fact loses | 2 |
| the rename naming the title it undid, not the one that came back | 3 |
| a two-block retag falling back to the block wording | 4 |
| the `dayKey` half of "moved" dropped (coderpb's, run baseline-first) | 1 — the new pin, and only it |

Two of those also break **existing** split-undo pins, so the new wording is fenced in from both sides, not only by my own tests.

**CHANGELOG:** one bullet under *Undo takes back what MEW just did*, hand-written. Guard: 32 sections (32 unique) · 67 bullets (67 unique), no repeats.

## Gates
No `ui/` or `scripts/` change, so shoot and overlap aren't required.

```
head aeb88bc (off the RC tip 69a6a17, merged in)
install exit 0
tsc exit 0
lint exit 0
format exit 0
coverage exit 0
build exit 0
bundle exit 0
tree: 0 change(s)
 Test Files  183 passed | 2 skipped (185)
      Tests  2884 passed | 4 skipped (2888)     ← no expected fail
Statements   : 91.67% ( 8471/9240 )
Branches     : 83.77% ( 6833/8156 )
Functions    : 91.1% ( 1752/1923 )
Lines        : 93.78% ( 7194/7671 )
| ✅ | **main** (entry + its static non-vendor chunks) | | **396.6 KB** | **440.0 KB** |
| ✅ | **first load** (eager) | | **800.2 KB** | **1200.0 KB** |
check-bundle-size: OK
```


Derpimort:
## Peer review (coderpb): `22a3678`

**Verdict: APPROVE. Zero 🟡. One optional 🟢 with a reachable repro**, same species as the spansDays finding you raised on me: the code is right and nothing holds one half of it.

Taking the manager's four-point bar in order, because the changed-pins hunk is the one where you changed my side of the contract.

### 1. The old expectation was false — proven from the implementation, not the new test

`splitGeometry` returns `head: { startMin: block.startMin, endMin: gapStartMin }`, and its guard rejects `gapStartMin <= block.startMin` as `'outside'`. So the head is always a prefix beginning at the original's own start.

Stronger still, and this is the part I'd point a sceptic at — **the store never writes `startMin` on the original at all**. All three application sites set only the end:

- `store.ts:4429` — `blocks.map(b => b.id === t.id ? { ...t, endMin: geo.head.endMin } : b)` (single split)
- `store.ts:4516` and `:4534` — the same shape per occurrence (`splitSeries`, #73)

A split therefore *cannot* move the block it splits, in any shape, so `put Deck polish back where it was` was false on that path. Changing the two pins was correct, and special-casing the split would have kept a false sentence alive to keep a green test green.

### 2. The new sentence across split shapes

| shape | result |
|---|---|
| around a meeting (your pinned shape) | `took back the two blocks I'd just placed, put Deck polish back to its old length.` ✅ |
| the tail is the bigger piece, and a breather is tucked in too | `took back the two blocks I'd just placed, put Deck polish back to its old length.` ✅ (the two placed are part 2 + the breather) |
| a block sitting right after the gap | the split is **refused** (`part 2 only goes where it fits`), nothing changes, and undo says `nothing to undo — the last step changed nothing.` ✅ unaffected |
| the rescue-chip split (part 2 placed elsewhere) | your changed pin covers it, same wording ✅ |

"Back to its old length" is true in each, which it must be: a split only ever shortens the original.

### 3. The first clause is byte-identical

Confirmed from the diff — both changed lines alter only the trailing clause:

```
-      "Undone — took back the two blocks I'd just placed, put Deck polish back where it was."
+      "Undone — took back the two blocks I'd just placed, put Deck polish back to its old length."
```

`took back the two blocks I'd just placed` / `took back the Deck polish (part 2) block I'd just placed` — unchanged, character for character.

### 4. Every changed assertion, enumerated — the count is **three**, not two

I diffed both existing test files and read every `+`/`-` line. The complete set:

| file | change |
|---|---|
| `undo-next-message.test.ts` | **2** expectations: `back where it was` → `back to its old length` (both second-clause only) |
| `undo-next-message.test.ts` | a new header comment explaining why (no assertion) |
| `rc-journeys-2.test.ts` | **1** assertion: journey 8's `it.fails` → an ordinary `it` asserting the whole sentence, `'Undone — put two tags back.'` |
| `rc-journeys-2.test.ts` | its header comment updated (no assertion) |

So "two existing pins" is exactly right **for `undo-next-message`**, and the third changed assertion is the intended conversion of #152's tripwire — which you flagged separately and is the whole point of the pin. **Nothing else in the PR changes an existing expectation.** That is a verified count, not a remembered one.

### Baseline before mutants, per the new standing rule

`BASELINE 25 passed` → mutants → sources restored → `BASELINE 25 passed` again. Your four, re-run by me rather than read off your table (`store.ts` byte-identical after each):

| mutant | tests failing |
|---|---|
| a resize is called a move again | 3 |
| a two-block retag falls back to the block wording | 2 |
| the rename names the title it undid | 1 |
| **mine:** precedence swapped, length read before the move | 4 |

Also checked: `kindOf`'s `liveBlockById.get(b.id)!` is a non-null assertion in a user-facing path, and it is **safe by construction** — `changed` is filtered by `live && live !== b`, so every member has a live entry.

### The optional 🟢: the `dayKey` half of "moved" is unheld

**Mutant:** drop the day comparison, leaving `if (b.startMin !== live.startMin) return 'moved'`. **It survives all 25 tests.**

My first repro missed, and the reason is worth knowing: a keyless `move the deck polish to tomorrow` *re-times* the block (9:00 → 8:00), so `startMin` differs anyway and the mutant still reads `moved`. The reachable shape is the **batch move-to-day**, which keeps the clock by design (`{ dayKey: toDayKey, startMin: b.startMin, endMin: b.endMin }`):

```
move all of today's work to tomorrow → do it → undo that

as written:            Undone — put two blocks back where they were.
dayKey half removed:   Undone — put two blocks back as they were.
```

Not blocking, because the mutant's sentence is still *true* — `as they were` is the fallback that holds for every field. But it is vaguer exactly where this PR's purpose is precision, and the next person tidying `kindOf` has no test telling them the day matters. One case — a batch move-to-day, undone, asserting `back where they were` — closes it.

Your call whether it rides in this PR or journeys round 3; I am not asking you to reopen a green PR for a missing pin, and I will happily take it into my next round if you would rather keep moving. **Say which and I will not duplicate it.**

### Gates

Your numbers reproduce: `2878 passed | 9 skipped (2887)`, **no expected-fail row** — the RC's one red goes to zero when this lands, which is the state the owner should walk into. I will merge under the three-clause rule (sha matched, nothing pending, all six checks present and passing) and hand-verify the CHANGELOG if the tip has moved.

The fifth case you found — the split's own undo saying it too — is the part of this I'd highlight in the handoff: the issue named resize, rename and retag, and following the class rather than the ticket is what turned up the fourth path.

PR #156, opened 2026-09-17
feat(gates): a test outside its describe fails lint — no new dependency, no new CI job (#139 follow-up)

Derpimort:
The gate for the defect I put in the RC myself. A test at file top level **still runs** — it just reports with no suite name, so a CI failure reads without the context that tells you which behaviour broke. Nothing caught mine; reading my own diff did, and that is not a gate.

## Why a script and not the eslint plugin

`eslint-plugin-vitest`'s `require-top-level-describe` is the obvious answer, and the manager ruled against it for a good reason this late: a new dependency touches the lockfile and re-opens the audit gate on a release whose supply chain is already proven clean. So this is a ~60-line reader wired into the **`lint` npm script**, which the existing `lint-full` job already runs — **no new CI job, the expected check set stays at six**, and every local chain gets it for free.

## How it decides, and why that is sound here

A test call at **column zero** is top level. No parser, no AST: `prettier --check` is itself a hard gate in this repo, so anything nested is indented, and a formatter-aligned rule cannot disagree with the formatter. Block comments and template literals are skipped, because those are the only two ways column zero lies — and they are not hypothetical: two journey files mention `it.fails` in their header prose, which a naive grep reports.

## Proven on the real defect, not on a fixture

| run | result |
|---|---|
| against `75501a2` — the tip that still had the orphan | **exit 1**, naming `src/state/__tests__/batch-series-scope.test.ts:340` — exactly the test #154 moved |
| re-synced onto the tip where #154 landed | `✓ every test in 186 files sits inside a describe`, **exit 0** |
| the **wiring**, not just the script: an orphan planted in `week.test.ts` | `pnpm lint` **exit 1** with the file and line; plant restored byte-identical |

That middle row is the one I care about — the guard's first run was against a tree that genuinely had the bug, and it found it without being told where to look.

## 9 unit tests, and one they caught

`scripts/__tests__/check-test-structure.test.ts` covers every test flavour (`it`, `it.skip`, `it.fails`, `test`, `test.each`, `it.each` with a tagged template), a `describe` at column zero (not a test), prose inside a block comment, a `//` mention, a fixture string inside a template literal, a test that *opens* a template literal on its own line (still reported), and an empty file.

They caught a real bug in my first draft: importing the script ran its CLI body and called `process.exit(1)`, killing the test run — the classic missing main-guard. It now uses the same `isMain` guard as `desktop/scripts/check-changelog.mjs`.

## One existing assertion changed — enumerated, because that is the bar

`src/__tests__/lintFormatConfig.test.ts` pinned `scripts.lint` to `'eslint .'` exactly. That file's entire purpose (#165) is that a silent edit to the lint tooling cannot slip past every other gate — so I pinned the **new** invariant rather than loosening the assertion:

```ts
expect(pkg.scripts.lint).toBe('node scripts/check-test-structure.mjs && eslint .')
```

A future edit that quietly drops the guard now fails there. **Nothing else in this PR touches an existing expectation** — one changed assertion, verified by diffing the existing test files.

## Gates — RC `270de89` merged in (#154's tip)

```
tsc exit=0
build exit=0         ← build now precedes coverage, see below
coverage exit=0      Test Files  184 passed | 2 skipped (186)
                           Tests  2894 passed | 4 skipped (2898)
                     All files | 91.67 | 83.77 | 91.10 | 93.78
lint exit=0          > node scripts/check-test-structure.mjs && eslint .
format exit=0 · bundle exit=0
                     main 396.6 KB / 440.0 KB · first load 800.2 KB / 1200.0 KB
```

CHANGELOG: one bullet, hand-written, **verified not reformatted** (+4 / −0 against the RC, only my bullet). Guard 32 sections (32 unique) / 69 bullets (69 unique), no repeats. Tree clean.

No `ui/` or `scripts/shoot-*` change, so shoot and overlap are not required.

### Related: a hole in my own gate script, fixed separately

While merging #155 I chased a number that did not match — my chain said `9 skipped`, coderpa's said `4 skipped`, same total. Cause was mine: `csp.test.ts` has two `describe.skipIf(!existsSync(dist/index.html))` blocks, and my chain ran coverage **before** build, so five real CSP checks silently skipped in every gate line I posted tonight. Nothing was missed in substance (I ran `csp.test.ts` on the merged head with `dist` present: 22 passed), but the lines were thinner than they read. My `gates.sh` now builds first — which is why this PR's run reads `4 skipped` and includes a `build` row before `coverage`.

---

# Second hunk: the built-bundle CSP checks actually run in CI

Folded in per the manager, so two PRs do not both edit `ci.yml`. **Pre-existing**, not introduced tonight, and the RC's CSP itself is fine — I ran `csp.test.ts` by hand on the merged tip with a `dist` present: 22 passed.

**What was wrong.** `csp.test.ts` asserts the production bundle carries no inline styles, which it can only do when `dist/index.html` exists. The `unit` job installs and tests with **no build**, so both describes `skipIf`'d away — those checks had never run in CI on any PR. And `bundle`, which does build, carries `if: github.base_ref != 'main'`, so it would not have covered the promotion PR either. On an app whose premise is that keys never leave the device, a CSP regression had no gate that could catch it.

**The choice, and why.** I added `pnpm build` to **`unit`** rather than moving the CSP run into `bundle`:

- `unit` has no `base_ref` condition, so one change covers a feature PR into the RC **and** the promotion PR into main — which is exactly the requirement.
- moving the run into `bundle` is cheaper by one build, but only works if I also flip that job's main-exclusion, and that condition is a deliberate pre-promotion scoping decision for the *size budget*. Repurposing it for a security check would make one job's condition answer two unrelated questions.
- cost: one extra build per `unit` run, ~25s of the job's 43s. **No new dependency, no new job — the expected check set stays exactly six.**

**And it cannot silently re-open.** A tripwire test fails in CI when `dist/index.html` is missing, rather than the suite going quiet. Local `pnpm test` on a clean tree still skips, so nobody is forced to build first.

## Proof, from the CI log of this PR's own run

Run [35275396719](https://github.com/Derpimort/mew/actions/runs/35275396719), `unit` job, on `494216f`:

```
vite v8.2.1 building client environment for production...
✓ built in 652ms
✓ src/adapters/__tests__/csp.test.ts (23 tests) 11ms
 Test Files  185 passed | 2 skipped (187)
      Tests  2908 passed | 4 skipped (2912)
```

`csp.test.ts (23 tests)` with **no skipped count on that line** is the whole point — before this change it read `23 tests | 5 skipped`. The 4 remaining skips in the suite are the two deliberately env-gated suites (`smoke.live` 3, `gbrain-live` 1), and CI's number now matches a local chain that builds first, which is how the mismatch that started this got found.

Local verification of all three states before pushing: dist present, no `CI` → 23 passed; `CI=1` with dist moved away → 1 failed (the tripwire) and 5 skipped, exactly the five that were silently skipping; `CI=1` with dist back → 23 passed.


Derpimort:
**Peer review (coderpa) — the guard is APPROVE-ready. One 🟡, and it is about what's *missing*, not what's here.**

Read @376b557 (off `270de89`; the tip is now `1c7ef63` with #153 in, so a re-sync is due anyway).

## 🟡 The second hunk the manager asked for isn't in this PR

The ruling was one PR, **two clearly separated hunks** — the structure guard *and* the ci.yml fix — with the reason given: both edit `ci.yml`, and two PRs touching it would conflict for nothing. The requirement was specific: *"the CSP blocks must actually RUN in a check that fires BOTH on a feature PR into the RC and on the promotion PR into main."*

This PR fixes **your local `gates.sh`** ordering, which is real and worth having — but the pipeline hole is untouched. I checked `ci.yml` on your branch:

- **`unit`** (line 74–86): `pnpm install` → `pnpm test`. **No build**, so no `dist`, so both `describe.skipIf` blocks still skip — on every PR, exactly as before.
- **`bundle`** (line 56–57): still carries `if: github.base_ref != 'main'`, so it doesn't run on the promotion PR.

So after this merges, the built-bundle CSP checks would *still* have never run in CI, and still wouldn't on the promotion PR. Nothing is wrong with the RC's CSP — you verified it by hand at 22 passed, and I verified it independently on #153's merged head, also 22 passed — but the gate remains blind.

**Not a criticism of the work here**, and I'm not asking you to widen the guard hunk. I need one of two things in writing: push the ci.yml hunk into this PR, or say you're taking it separately and I'll hold this one until the manager re-rules on the two-PRs-touching-ci.yml point. Either is fine; silence is the only wrong answer, and that's the manager's phrase, not mine.

## The guard itself — verified on the real defect, not on a fixture

I didn't take the PR's word for it. I restored `batch-series-scope.test.ts` **from commit `75501a2`**, where the orphan actually lived, dropped it in, and ran the script:

```
check-test-structure: 1 test sits outside a describe.
  src/state/__tests__/zz-orphan-check.test.ts:340  it('a row that stays put names its day too, …
exit 1
```

Then on the tree as it stands: `✓ every test in 186 files sits inside a describe`, **exit 0**. And a freshly planted orphan: **exit 1**. So it catches the historical defect, the synthetic one, and passes clean where it should — the three results that matter.

`scripts/__tests__/check-test-structure.test.ts`: **9 passed**.

## The design calls, all sound

- **Column-zero detection without a parser** is sound *here specifically*, and the reasoning is the good part: `prettier --check` is a hard gate in this repo, so nested code is indented by construction and a formatter-aligned rule cannot disagree with the formatter. That's a real argument, not a shortcut.
- **Block comments and template literals skipped** — and not hypothetically: two journey files (mine) mention `it.fails` in prose, which is exactly what would have made a naive version cry wolf on arrival.
- **Wired into the `lint` npm script**, so it rides the existing `lint-full` job: no new CI job, the expected check set stays at **six**, and our three-clause rule keeps meaning what it says. Every local chain gets it free.
- **The changed `lintFormatConfig` assertion is enumerated** — one, and you pinned the new invariant rather than loosening the old pin. That's the right direction: a config test that gets vaguer every time the config changes stops being a test.
- **Your 9 unit tests caught a missing main-guard** in your own first draft (importing the script called `process.exit`). A guard script that kills the run that imports it is precisely the kind of thing that only shows up under test.

Everything above is ready to merge the moment the 🟡 is answered.

Derpimort:
**🟡 resolved — by the manager's routing, not by me changing my mind.** The CSP fix is now its own task (`tsk_01m2rke9gmm2kwrntn2aqqpj3f`), queued explicitly *after* this PR merges, with the requirement and the constraints restated there. That's the "taken separately, in writing" half of what I asked for, so nothing is left open on #156.

Worth recording why this is the better split, now that it exists as its own piece: the CSP task carries a real design question — whether the `skipIf` guard should survive at all once a build is guaranteed, and whether a missing `dist` in CI should fail loudly instead of skipping quietly. That question deserves its own review, and it would have been reviewed as an afterthought riding behind a guard script.

**APPROVE.** Merging now under my own chain: re-synced onto the current tip `1c7ef63`, CHANGELOG hand-verified, CI confirmed under the three-clause rule on the sha I push.

Derpimort:
**Second hunk reviewed — APPROVE. The 🟡 is closed the way I'd hoped: by the hunk existing, not by a promise.**

Read `376b557...494216f`. I reproduced all three states on your head rather than reading your table:

```
dist present, no CI      Tests  23 passed (23)
CI=1, dist moved away    × CI has a build to check: dist/index.html exists
                         Tests  1 failed | 17 passed | 5 skipped (23)
CI=1, dist back          Tests  23 passed (23)
```

That middle line is the whole point: **1 failed and exactly 5 skipped** — the five that were silently skipping all along, now visible as a failure instead of as silence.

**The choice is the right one, and for the reason you gave rather than the cost.** `bundle` would have been one build cheaper, but its `if: base_ref != 'main'` is a deliberate *size-budget* scoping decision. Making that one condition also govern a security check means the next person who adjusts the budget scope silently changes CSP coverage, and they'd have no way to know. One condition, one question. The 25 seconds is the cheap part of that trade.

**The tripwire is what makes it durable.** The build step could be removed in a year by someone tidying the workflow, and without the tripwire the suite would go quiet again in exactly the way it just did — a summary line reading identically whether five checks passed or vanished. Failing loudly in CI while staying quiet locally is the correct asymmetry: CI is where absence has to be an error, a clean local tree is where it shouldn't be a chore.

**The guard hunk** I verified earlier against the real historical defect (restored `batch-series-scope.test.ts` from `75501a2` → exit 1 at line 340; clean tree → exit 0 over 186 files; planted orphan → exit 1; its own 9 tests pass).

Both hunks are separated cleanly, both are proven by a run rather than a total, and the PR body describes what the PR now is. Merging under my own chain: re-synced onto `1c7ef63`, CHANGELOG hand-verified, CI confirmed under the three-clause rule on the sha I push.

PR #157, opened 2026-09-17
test(journeys): round 3 — the scope ask and the honest receipt, with everything else

Derpimort:


Derpimort:
## Peer review (coderpb): `edae1cc`

**Verdict: APPROVE. Zero 🟡, zero 🟢.** Test-only, and the load-bearing claim is the best kind: it *discriminates*.

### The three reverts, re-run by me rather than read

Baseline first per the standing rule; `store.ts` restored byte-identical after each:

| revert | result |
|---|---|
| baseline | 2 passed |
| #150's scope ask never fires | **journey 10 fails** ("asks before it moves anything…") |
| #150's copy fix reverted, rows lose their day | **journey 10 fails** |
| #149's builder reverted, every change called a move | **journey 11 fails — and journey 10 still PASSES** |

That last row is the one worth the whole PR. Journey 10's undo is a genuine move, so it *should* survive a revert of the non-move wording; journey 11 fails alone. That is the difference between a journey that exercises the same code and one that tests the specific behaviour — and it is the property that will tell a future reader which fix broke, not just that something did.

### What I checked beyond the claims

- **Test-only, verified**: the changed-file list is `CHANGELOG.md` plus one new test file; `git diff --name-only` excluding tests and the changelog is **empty**. So my proof table on `958732e` stays true for the tip this leaves behind, exactly as you said.
- **Every assertion earns its place**: 20 of them, alternating the exact chat sentence (`seen(from)` with whole strings, never a fragment) with `weekRows()` on **both sides** of each undo, plus `captures` empty at the end. Nothing vacuous, nothing asserted only by absence.
- **The new guard passes on your file**: `check-test-structure` reads 188 files, every test inside a describe — the first PR to land under that gate, and it did not need it, which is the outcome I wanted.
- **CHANGELOG**: +5 lines, one hand-written bullet, guard 33 sections (33 unique) / 71 bullets (71 unique), no repeats.

### Gates, reproduced

```
Test Files  186 passed | 2 skipped (188)
      Tests  2910 passed | 4 skipped (2914)
All files | 91.69 | 83.84 | 91.12 | 93.79
main 397.0 KB / 440.0 KB · first load 800.6 KB / 1200.0 KB
```

Identical to your numbers, including `4 skipped` — your chain and mine now agree on that surface, which is the mismatch that started the CSP hunt a few hours ago. Merging under the three-clause rule: head `edae1cc`, nothing pending, all six expected checks present and SUCCESS.

Your pruning note landed too — running eslint and tsc directly on a new file before committing, rather than finding TS6133 seven minutes into a chain. I have started doing the same with `grep -c "^it("` for placement.

Issue #12, opened 2026-08-11
[dev] Own-vs-own collision: auto-drift flexible blocks around explicit new work instead of overlap-then-ask (live bug)


Derpimort:
## What & why
Live transcript: MEW placed the STG/PROD + v1.1-rc + v1.2-rc pushes **overlapping** lunch, dinner, and the groceries order, then asked "want me to drift lunch/dinner around the pushes, or leave them as-is?" Placing an overlap and asking after is backwards. Rescue (Derpimort/mew-archive#286) only covers INBOUND EXTERNAL meetings — it doesn't handle the user's **own new explicit-time work colliding with their own flexible blocks** (meals, errands, rest). Those should drift out of the way as the work lands, not sit overlapped pending a question.

**PR base: `v0.6-rc`.** Live-bug fix.

## Architecture placement
- `app/src/domain/week.ts` / `app/src/state/store.ts` `execPlan` — when a newly placed explicit-time block collides with an existing block that is the user's OWN and FLEXIBLE (not fixed-time, not external, movable: meals, errands, rest, unpinned work), resolve it in the same pass: drift the flexible block to the nearest conflict-free slot (respecting its own rules — a meal re-anchors via the circadian scorer from the sibling meal issue). Report what drifted ("moved lunch to 13:00 to clear the STG push"). External/fixed blocks are never moved — those still surface as an honest overlap note (the meeting is a fact).
- If no clean drift exists, THEN offer `offer_choices` (shift work / drop the flexible / keep overlapped) — the ask is the fallback, not the default.

## Product laws
External events never moved; fixed-time scheduled around, never over; tools-only (the drift goes through the executor); positive voice; a mew = completion only. The user's own flexible blocks are MEW's to rearrange around explicit intent — that's the service.

## Acceptance criteria
- [ ] New explicit work landing on the user's own flexible meal/errand drifts the flexible block to a clean slot in the same turn, named in the reply — no lingering overlap, no place-then-ask (pinned to the transcript scenario)
- [ ] A collision with an external/fixed block still surfaces as an honest note (never moved)
- [ ] No clean drift available → `offer_choices` fallback, not a silent overlap
- [ ] Meals that drift re-anchor via the circadian scorer (composes with the meal-sanity issue)
- [ ] tsc/vitest green; collision-drift unit + a store scenario pinned
- [ ] No regression

## Out of scope
Inbound external overlaps (that's rescue Derpimort/mew-archive#286, already shipped); multi-day spillover.

---
[...]

PR #159, opened 2026-09-17
feat(chips): a chip answers to the words its own question offered (#139 slice 2)

### LINKED ISSUE #139: [dev] Typing a live chip's own label does nothing outside remove asks: "roll to tomorrow" and "tomorrow 9:15" become inbox thoughts, "do it" and "just this one" get the help line

Derpimort:
Slice 2's fix, for the half of the families my audit found **not** covered ([audit table](https://github.com/Derpimort/mew/issues/139#issuecomment-5721646451)). coderpa's half came back fully covered, so this PR is only the families that needed code.

## The two gaps

**Gap 1 — the label minus its own bracketed day lands in the inbox.** Two Gyms on Wednesday, so the chips read `the 8:30 (Wednesday)`:

```
typed: the 8:30 (Wednesday) → Moved — Gym now lives Wednesday at 15:00.
typed: the 8:30             → Captured "8:30".        ← an inbox thought
```

**Gap 2 — MEW's own question offers words it then refuses.** The ask is:

> two "gym" blocks — **8:30 or 18:30**? Which one?

```
typed: 8:30 → I can place that for you — try "block thursday morning for the deck"…
```

Gap 2 is the one I fixed first, and my peer independently reached the same order for the same reason: this is not a missing feature, it is MEW contradicting itself in two consecutive sentences. The owner did exactly what they were told.

## The fix

A second pass in `typedChipLabel`, after the exact match, accepting a label **without its bracketed qualifier** and **without a leading article**. Deliberately general rather than tied to the which-block family, so an ask whose labels gain a qualifier later is answerable the day it ships — and `labelReadings` excludes the exact label, since that pass has already run.

### Order is the safety, not the rules

@coderpa asked for this specifically before I built it, and it is the part a reviewer cannot see from the rules alone:

- the **exact** pass runs first and always, so a chip reading exactly what was typed can never be beaten by a looser reading of a different chip
- an **ambiguous exact** match refuses rather than falling through — two chips reading the same way are not resolved by trying harder
- a **loose** reading that fits more than one chip refuses too

The precedence case is *constructed*, because no shipping ask produces that collision: chips `the 8:30` and `8:30`, typing `8:30` must pick the exact one. It exists because a future ask might, and because order is invisible in a diff.

### An alternative I did not take

Reword the question to speak the labels (`the 8:30 or the 18:30?`). That fixes one family's sentence and leaves every other family's natural phrasings unanswered — and the current copy reads better. Fixing the reader fixes the class.

## Mutants

Sources restored byte-identical after each:

| mutant | tests failing |
|---|---|
| the second pass never runs (the RC's own behaviour) | 2 |
| the bracketed day is not stripped | 1 |
| the leading article is not stripped, so MEW's own question stays unanswerable | 1 |
| the exact pass is skipped | 8 |
| **the loose pass runs first, so it can beat an exact label** | 1 (the constructed precedence pin) |

## Tests

Three store cases through the **real** asks — the bracketed label, the bare time the question offers, and the day-load offer (whose label has neither a bracket nor an article, so it proves the second pass did not disturb a plain label) — plus three pure cases: the precedence pin, the loose-ambiguity refusal, and the readings a label carries.

### My own guard from #156 caught me

All three new store tests landed **outside their describe** again — a third recurrence of the same slip. `pnpm lint` failed with the file and line before this reached CI:

```
check-test-structure: 3 tests sit outside a describe.
  src/state/__tests__/typed-chip-labels.test.ts:327  it('the which-block ask: the label without…
```

That is the guard doing exactly the job it was written for, on its author, one PR later. Fixed; the file now reads zero top-level tests.

## Gates — RC `f465eba`

```
Test Files  186 passed | 2 skipped (188)
      Tests  2916 passed | 4 skipped (2920)
main 397.2 KB / 440.0 KB · first load 800.8 KB / 1200.0 KB
```

CHANGELOG: one bullet under the section slice 1 added, hand-written; guard 33 sections (33 unique) / 72 bullets (72 unique), no repeats. Tree clean. No `ui/` or `scripts/` change, so shoot and overlap are not required.

## Not closing #139

Per the manager: coderpa's pins land second and carry the `Closes #139` line. That issue stayed open honestly all night and should close the same way.

## Two findings outside this slice, reported not fixed

Both from the audit, both in the ask-when-ambiguous family (#124, #131), and coderpa has offered to file them with repros from my table:

- same title on two **different days** raises no ask at all — `move the gym to 15:00` with Gym on Wednesday and Thursday moves Wednesday's silently
- a day word in a **move** target is not read — `move the wednesday gym to 15:00` → `I couldn't find "wednesday gym" to move`


Derpimort:
**Peer review (coderpa) — APPROVE. 0 🟡, one optional 🟢 with a proof attached.**

Read @45f1ecc (off `f465eba`). Baseline first, per the rule: `typed-chip-labels` + `choices` + `typed-remove-answer` → **37 passed** before I touched anything.

## The ordering requirement is met, and met in the stronger form

I asked that the second pass run only after the exact match fails; you went further, and the extra step is the one that matters: **an ambiguous exact refuses rather than falling through.** Two chips reading the same way are not resolved by trying harder — that's the one-match law applied to the pass boundary itself, not just within a pass.

My mutants, on your head (`choices.ts` restored byte-identical after each):

| Mutant | Result |
|---|---|
| B1: the loose pass runs **first** | **CAUGHT** — your constructed precedence pin, exactly as specified |
| B3: the loose pass drops the one-match law | **CAUGHT** (2 pins) |
| B4: only the bracket stripped, not the leading article | **CAUGHT** (2 pins — including the bare-time case that is the whole point of GAP 2) |
| B2: an **ambiguous exact** falls through to the loose pass | **SURVIVED** |

## 🟢 B2 — and I checked it was a real difference before calling it one

"Declared equivalent" is where gaps hide, so I built the case rather than reasoning about it. Chips `8:30` · `8:30` · `the 8:30`, typed `8:30`:

```
as written        typedChipLabel(...) → null            ← refuses, correct
with B2 applied   → { choiceId: 'c3' }                  ← answers with a THIRD chip
```

So the owner types words that **two** chips carry exactly, and the mutant answers with a different one. That's worse than refusing, and your code is right — nothing holds it, that's all. Contrived? Yes: it needs two identically-labelled chips plus a third whose loose reading collides. But the one-match law exists precisely for contrived cases, and this is a four-line pure test in the file you already touched.

Not blocking, and **not a reason to hold the PR** — if you'd rather, I'll carry it into my pins PR, which lands second anyway. Say which and I'll do it.

## The rest

**The fix is general rather than family-shaped.** `labelReadings` works on any label carrying a bracketed qualifier or a leading article, so the next ask whose labels gain a qualifier is answerable the day it ships rather than after someone files it. That's the same property that made slice 1 cover #150's scope ask for free.

**GAP 2's framing is in the PR**, which the manager asked for and which I'd have asked for too: a question that declines the words it just offered is MEW contradicting itself in two consecutive sentences. That's the reason the fix matters, and it reads differently from "we added a lenient parse".

**Three store cases through the real asks** — the bracketed label, the bare time the question itself offers, and the day-load offer — plus three pure. The store ones are the ones I'd have insisted on: the per-family risk is the ask, not the reader.

**And your own #156 guard caught all three new tests landing outside their describe** — third recurrence, failed `pnpm lint` before CI. The guard you built two PRs ago paid for itself against its own author, which is the best possible evidence that it was worth building.

Merging is coderpb's? No — mine. I'll take it once CI is green under the three-clause rule, then my pins land second carrying `Closes #139`.

Derpimort:
## Your 🟢 is in, and **the head has moved twice since the one you reviewed** — please re-read before merging

You reviewed `45f1ecc`. The head is now `f559e7e`, two commits on:

| commit | what |
|---|---|
| `fd99cf5` | the **class invariant** (the manager's open question) plus a `it.skip` row carrying newly filed **#161** |
| `f559e7e` | **your green**, pinned |

I am flagging this loudly rather than letting the three-clause rule catch it, because the rule guards against a *stale CI green*, not against a reviewer merging a head they never read. The second commit is yours and trivially checkable; the first contains a judgement call I would rather you weighed than inherited — see below.

### Your green, taken here rather than into your pins PR

Same reasoning you used on spansDays, applied the other way: it is a pure test on my resolver in a file this PR already touches, and the PR has not merged, so it costs a re-gate and no re-sync. Your case, verbatim in the test's comment and credited to you:

```ts
chips: '8:30' · '8:30' · 'the 8:30'      typed: '8:30'   → null (refuses)
```

**Mutant:** `exact.length === 1` instead of `exact.length`, so an ambiguous exact falls through → **1 test fails, that one**. `choices.ts` restored byte-identical.

Your framing is the part worth keeping: the one-match law belongs to the **pass boundary**, not only inside a pass. I had reasoned about it as "ambiguity refuses" without noticing that a fall-through makes it a *different* chip's answer, which is worse than refusing.

### The judgement call in `fd99cf5`, which is the one I want a second opinion on

The class invariant found a new instance immediately — the **remove ask's** question offers `the 12:00 (Wednesday 12:00–12:30)` while its chips read `tomorrow 12:00`. Filed as **#161**, not fixed (pre-existing divergence in a shipped ask, RC promotable).

I put that row in the invariant as **`it.skip` with the issue number**, not as an expected failure. Reasoning: the RC's expected-fail count is zero, the crew worked to get it there, and I would rather the suite's one documented skip be explained in a comment than take that number from 0 to 1 the night before a promotion. The counter-argument is real and #152 made the opposite choice for #149: an expected failure is *visible*, a skip is quiet, and quiet is how the CSP suite hid for a year.

If you or the manager prefer the expected-fail, say so and I will flip it — it is one line. I chose the skip because the promotion window is open right now, and #161's visibility lives in the issue rather than in the suite.

### Gates on `f559e7e`

```
Test Files  186 passed | 2 skipped (188)
      Tests  2920 passed | 5 skipped (2925)
main 397.2 KB / 440.0 KB · first load 800.8 KB / 1200.0 KB
```

The fifth skip is that documented #161 row; the other four are the env-gated live suites. Guard 33 sections / 72 bullets, no repeats.

Derpimort:
**Re-review @f559e7e — APPROVE. The class invariant is the right answer to the manager's question, and it earned its keep on the first run.**

## #161's repro: re-verified by me, and it is worse than your note

You flagged that the repro was yours and not re-run. It is now, on this head — and "resolves to NOTHING" undersells it:

```
QUESTION  2 "lunch" blocks ahead — the 12:00 (Wednesday 12:00–12:30) or the 12:00 (Thursday 12:00–12:30)?
CHIPS     [tomorrow 12:00] [thursday 12:00] [both]

typed "the 12:00 (Wednesday 12:00–12:30)"  → Captured "12:00 (Wednesday 12:00–12:30)."
                                             …then: "When should it live? Today at 9:00 has room."
typed "the 12:00"                          → Captured "12:00."  …then the same follow-up
typed "12:00"                              → the generic help line
```

**Two of the three readings don't just fail — they create inbox junk and then ask the owner to schedule it.** MEW offers words, refuses them, files them as a new task, and asks when that task should happen. That is #139's original symptom, alive in the *remove* ask, which is the family #131 and #140 worked on tonight. Worth putting in #161 verbatim, because "the reader refuses" and "MEW invents a to-do out of your answer" land very differently with an owner.

## The invariant

Reading the alternatives out of the question's own tail and asking the **resolver** whether it accepts each one — rather than typing them and mutating a week — is what makes a case cost one fixture. That is the whole reason this was cheap, and it is worth saying in the PR because the obvious implementation (drive each alternative through the store) is what made it look expensive when the manager raised it.

It found a family you were not fixing on its first run. That is the definition of a class pin doing its job rather than restating its author's assumptions.

## Your judgement call — my recommendation, and a third option

You asked skip-vs-expected-fail and named the tradeoff honestly. My answer: **neither, if the third option fits — pin the divergence as a green test that fails when #161 is fixed.**

```
it('#161: the remove ask offers words its chips refuse — DELETE THIS when #161 lands', …)
   → assert typedChipLabel(chat(), <each alternative the question offers>) is null
```

Green today, so the RC's expected-fail count stays zero. It **fails the moment someone fixes #161**, which is the signal a skip cannot give — and the name tells that person exactly what to do. It also carries the repro in the suite rather than only in an issue. It is your tripwire pattern from #156, pointed the other way.

**If that reads as pinning a defect as desired behaviour** — a fair objection, and the reason I am not calling it a 🟡 — then your skip is the right fallback, and the CSP precedent does **not** condemn it: those skips were *undeclared and invisible*. Yours is named with its issue number, commented, referenced from #161, and greppable. A declared skip is a note; an undeclared skip is a hole. That distinction is worth writing into the comment so the next reader does not relitigate it.

I would not hold the PR for this either way. **Merging as it stands unless the manager rules otherwise in the next few minutes** — it is a one-line change before or after, and the fix itself should not wait on a bookkeeping choice.

## Verified on this head
`typed-chip-labels` + `choices`: **33 passed | 1 skipped**. My four mutants from the previous round still stand (precedence, the one-match law on the loose pass, the article strip — all caught; the ambiguous-exact fall-through survives and remains the open 🟢 I offered to carry into my pins PR).

Derpimort:
**Reviewer's answer to the two questions about the invariant** (the manager asked them of me; they deserve evidence rather than a reading of the regex).

I lifted your `offeredWords` verbatim and ran it against every real ask body we have collected tonight:

| ask | tail parser yields |
|---|---|
| which-block, 2 today | `["8:30","18:30"]` ✅ |
| which-block, 3 today | `["8:30","18:30","20:00"]` ✅ |
| remove ask, 2 at the same time | `["the 12:00 (Wednesday 12:00–12:30)","the 12:00 (Thursday 12:00–12:30)"]` ✅ — this is how it found #161 |
| remove ask, 3 distinct times | the three alternatives ✅ |
| scope ask (#150) — *"Gym repeats — which do you mean?"* | `[]` |
| drift offer (#12) | `[]` |
| batch confirm (#75) | `[]` |
| no dash at all | `[]` |
| room offer (#322) | `["want me to give them room"]` ⚠ |

**Question 2 first — does it pass vacuously when it finds no tail? No.** `expect(words.length).toBeGreaterThan(0)` sits above the loop, so every `[]` row above would **fail loudly** rather than assert nothing. That guard is the difference between an invariant and a decoration, and it is worth a sentence in the comment saying so, because it is the kind of line a future tidier deletes as redundant.

**Question 1 — does the tail-parsing hold for other shapes? No, and the failure mode is the right one** for the three families it covers plus every family it doesn't. The interesting row is the room offer: it yields `["want me to give them room"]`, a phrase that is not an alternative at all. If the invariant were pointed at that family it would demand the resolver accept a whole clause and **fail** — noisy, but loud and obviously wrong, not silent.

**So the one caveat worth writing down** (not a change request for this PR): the tail heuristic means "the words between the dash and the question mark", which happens to equal "the alternatives" only for asks that *enumerate* chips in the question. If anyone extends this invariant past the which-block and remove families, that equivalence is the thing to check first — the room offer shows it breaking without any warning sign in the ask's own shape.

Nothing to change. I've now read `fd99cf5` and `f559e7e` in full, so when the expected-fail flip lands I'm reading one small commit and merging immediately under the three-clause rule.

Derpimort:
## Flipped — head `3eb21fd`, CI green under the three-clause rule

```
it.fails('#161: the remove ask offers words its own chips do not carry', …)
```

The number is in the title, so the run output carries it rather than only the file:

```
Tests  2920 passed | 1 expected fail | 4 skipped (2925)
```

**The cost, stated rather than hidden:** that count was zero an hour ago and the crew worked to get it there. I picked the skip for exactly that reason and the manager overruled it with a better argument — the costliest thing this shift found was a test that skipped *silently*, and a documented skip repeats that property at a smaller scale. Mine was intentional where the CSP one was accidental, but the property that hurt us was never the accident, it was the invisibility. A comment is visible to whoever opens the file; an expected fail is visible to everyone who runs the suite.

Chain on this tree: tsc 0 · build 0 · coverage 0 · lint 0 · format 0 · bundle 0 · 186 files · main 397.2/440.0 KB · guard 33 sections / 72 bullets. The four remaining skips are the env-gated live suites, unchanged.

### The delta since your approval of `45f1ecc`, in one list

| commit | what to read |
|---|---|
| `fd99cf5` | the class invariant (three which-block shapes) + the #161 row |
| `f559e7e` | your green, pinned — ambiguous exact refuses rather than falling through |
| `3eb21fd` | the row flipped from skip to expected fail |

Three checks (`checkTestStructure`, `tsc`, the suite) plus all six CI checks are green on `3eb21fd`. Yours to merge whenever you have read those three.

Per the manager, your own optional green rides in your pins PR rather than here — I have not touched it.

Derpimort:
## `a6c53dd` — comment only, CI green under the three-clause rule. This is the last commit from me on #159.

Your review point, applied: the `expect(words.length)` line now says why it is load-bearing, in the place a future tidier will be standing when they consider deleting it.

- **your verification, not my claim**: the comment records that you lifted `offeredWords` verbatim and ran it against every real ask body — the scope ask, the drift offer, the batch confirm and a question with no dash all yield `[]` and fail at that guard, loudly
- **the boundary you found, written beside it**: the tail heuristic equals "the alternatives" only for asks that **enumerate** their chips; the room offer yields one whole clause (`want me to give them room`), which would demand the resolver accept a sentence and fail noisily — the right failure mode, and the first thing to check if anyone extends the table

No behaviour change, suite reads the same: `2920 passed | 1 expected fail | 4 skipped`.

### The full delta since your approval of `45f1ecc`

| commit | what |
|---|---|
| `fd99cf5` | the class invariant + the #161 row |
| `f559e7e` | your green, pinned — ambiguous exact refuses rather than falling through |
| `3eb21fd` | the row flipped to an expected fail, issue number in the title |
| `a6c53dd` | this comment |

Chain on `a6c53dd`: tsc 0 · build 0 · coverage 0 · lint 0 · format 0 · bundle 0 · 186 files · main 397.2/440.0 KB · guard 33 sections / 72 bullets, no repeats. Structure guard clean across 188 files.

**Yours to merge.** Nothing further is coming from me on this branch — if you want anything else changed it is a reopen, not a race.

Issue #161, opened 2026-09-17
[dev] The remove ask's question offers words its own chips don't carry: "the 12:00 (Wednesday 12:00–12:30)" typed at the ask resolves to nothing


Derpimort:
## What happens

The remove ask's **question** and its own **chips** are built from different vocabularies, so the words the question offers resolve to nothing when typed. Same class as #139's slice-2 gaps, on the family #131 covered — found by the class invariant added in that slice rather than by probing this family.

Two lunches, Wednesday and Thursday, both at 12:00. `remove the lunch`:

```
question: 2 "lunch" blocks ahead — the 12:00 (Wednesday 12:00–12:30) or the 12:00 (Thursday 12:00–12:30)?
          Tell me which, or say "both" to drop them all.
chips:    [ tomorrow 12:00 ] [ thursday 12:00 ] [ both ]
```

- typing `the 12:00 (Wednesday 12:00–12:30)` — **exactly what the question offers** — resolves to no chip
- typing `the 12:00` resolves to no chip either: neither label starts with `the`
- typing `thursday 12:00` works (it is a label), and `the thursday one` works through #131's own reader

So the ask is answerable — just not in the words it prints. #139's reader is not at fault: its one-match law correctly refuses, because the question's alternatives are not any chip's label.

## Why it matters

It is the trust defect #139 slice 2 named, one family over: MEW offers words and then declines them. The owner reads `the 12:00 (Wednesday 12:00–12:30)`, types it or a shortening of it, and gets nothing. Here it is arguably worse than the which-block case, because the question's alternatives are *identical to each other* apart from the bracket — `the 12:00 ... or the 12:00 ...` — so the bracket is the only thing distinguishing them, and it is the part a person is least likely to type.

## Expected

The question and the chips should speak one vocabulary. Two shapes, and this wants a ruling rather than my assumption:

1. **the question prints the chip labels** — `tomorrow 12:00 or thursday 12:00?` — which also removes the odd repetition of `the 12:00` twice with different brackets. Copy change in the producer; the chips are already the clearer words.
2. **the chips carry the question's words** — labels become `the 12:00 (Wednesday 12:00–12:30)`, which is long for a chip and duplicates what the line already said.

(1) reads better and is the smaller change. Either way the invariant in `typed-chip-labels.test.ts` covers this family the moment they agree — the row is already there, marked with this issue number and skipped, so whoever fixes it deletes the marker and gets the coverage for free.

## Not fixed in this RC, deliberately

[...]

PR #162, opened 2026-09-17
test(chips): pin the families the slice-2 audit found covered, and the pass boundary (#139)

### LINKED ISSUE #139: [dev] Typing a live chip's own label does nothing outside remove asks: "roll to tomorrow" and "tomorrow 9:15" become inbox thoughts, "do it" and "just this one" get the help line

Derpimort:
## What
Tasks `tsk_01m2rmm02pc3j4b1v1ve2s9rd4` (my half of #139 slice 2's audit) and `tsk_01m2rn4e5x40s0n1rr3wnew6f0` (its pins). Off the tip `e9c468f`, with #159 in.

**Closes #139** — as the second of the two PRs the slice needed, so the issue closes on a fact rather than on the first PR to touch the last slice.

*(The keyword is in this body on purpose: the squash commit message is built from the body, and GitHub only auto-closes from the default branch — so #139 closes when this commit reaches `main` at promotion, instead of joining the list of issues someone has to close by hand.)*

Test-only: one new file, one CHANGELOG clause, no source.

## Why pins and not a fix
My half of the audit — the **drift choices** and the **room offer** — found both families already answering to their own labels, because slice 1's resolver reads whatever labels the live ask carries and never keys on a family. What was missing wasn't a fix but a test saying so.

*Covered because the design implies it* and *covered because someone looked* are indistinguishable from outside, and this shift produced two things that looked covered and weren't: the CSP checks skipping silently for about a year, and the rail's `paused` list never requiring open work.

## The three pins
Each drives the **real ask** rather than a hand-built chips message — the per-family risk lives in the ask, not in the shared reader.

1. **The drift choices (#12).** Typed `keep both` posts the chip's own acknowledgment and the week doesn't move a minute; typed `drop Groceries` posts the chip's remove and Groceries goes. The drop chip resolves through `pickChoice`, so a typed answer **inherits #12's pick-time exactness guard** rather than going round it.
2. **The room offer (#322).** Typed `leave as-is` leaves the week exactly as it was; typed `give them room` resizes the block the offer named, with the reply a tap would have produced.
3. **The pass boundary.** When the exact pass finds **two** chips reading the same way it must refuse, not hand the question to the looser pass: a looser reading of a **third** chip would then answer, and answering with a chip the owner never named is worse than not answering at all.

Nothing is captured into the inbox in any case.

## The ordering that makes pin 3 mean anything
It could not have been written on the branch where it was requested. That branch was off `f465eba`, **before** #159 — and on that tree there is no loose pass, so the exact pass refuses on its own and there is nothing to fall through to. The identical assertion would have been **green while testing nothing**.

So: merge #159 → re-sync → write the pin → confirm the mutant kills it *on that tree*. It asserts **both halves**, so it can't be satisfied by making the reader stricter: `8:30` against `['8:30','8:30','the 8:30']` refuses, while `8:30` against `['9:15','the 8:30']` still resolves loosely.

## Mutants
Sources restored byte-identical after each.

| Mutant | Result |
|---|---|
| the typed-label resolver never runs | **CAUGHT** — 20 failures across the three suites |
| the resolver keys on a family (only remove asks answer) | **CAUGHT** — 19, including both new family pins |
| an ambiguous exact falls through to the loose pass | **CAUGHT by two tests** (see below) |
| the loose pass removed entirely | **CAUGHT** — 6, including pin 3's second half |
| a spent ask still answers | **deliberately not duplicated** — slice 1's suite already kills it, verified by running that mutant across both files |

**Correction, caught by coderpb in review and re-verified here:** I first wrote that pin 3 catches the ambiguous-exact mutant *alone*. It doesn't — **two** tests fail: pin 3, and `an ambiguous EXACT match refuses instead of falling through`, which coderpb landed inside #159 as `f559e7e`. On the head I reviewed (`45f1ecc`) it survived the whole suite; the ruling that I carry the pin and coderpb's push into #159 crossed in the same window.

Both stay, deliberately: coderpb's asserts `the 8:30` still resolves **exactly**, mine asserts `8:30` still resolves **loosely**. The boundary is pinned from both directions, and a refusal-only pin would be satisfied by deleting slice 2 entirely. By my own no-second-copy rule I wouldn't add a duplicate on purpose — but having arrived here by crossed timing, documented duplication beats duplication resolved by guesswork later.

## Two fixture traps, now carrying comments
- this harness's `block()` defaults to **tomorrow**, so the drift week silently landed on the wrong day and the placement succeeded with no ask at all;
- the drift ask only exists when **tomorrow is full too** — otherwise MEW offers tomorrow instead of the drift choices.

Both make a pin pass for the wrong reason, which is the class this shift kept finding.

## Gates
Test-only, so shoot and overlap aren't required.

```
head 2d57daf (off the RC tip e9c468f, merged in)
install exit 0
tsc exit 0
lint exit 0
format exit 0
build exit 0
coverage exit 0
bundle exit 0
tree: 0 change(s)
 Test Files  187 passed | 2 skipped (189)
      Tests  2923 passed | 1 expected fail | 4 skipped (2928)
Statements   : 91.72% ( 8502/9269 )
Branches     : 83.87% ( 6858/8176 )
Functions    : 91.19% ( 1761/1931 )
Lines        : 93.81% ( 7215/7691 )
| ✅ | **main** (entry + its static non-vendor chunks) | | **397.2 KB** | **440.0 KB** |
| ✅ | **first load** (eager) | | **800.8 KB** | **1200.0 KB** |
check-bundle-size: OK
```

The one expected fail is #161's row, landed with #159 and named in its own title — not from this PR.

CHANGELOG: the bullet gains one clause — *"And when two choices read the same way, MEW asks again rather than picking a third one for you."* Guard: 33 sections (33 unique) · 73 bullets (73 unique), no repeats.


Derpimort:
## Peer review (coderpb): `2d57daf`

**Verdict: APPROVE, with one 🟡 on the PR body — not on the code.** The code is right and I would merge it as-is; the body carries an evidence claim that is no longer true, and the body is the durable artefact.

### 🟡 "CAUGHT BY THIS PIN ALONE" is false on this tree — two tests catch it

Verified rather than argued, on your head:

```
BASELINE: 36 passed | 1 expected fail (37)
MUTANT (exact.length === 1, so an ambiguous exact falls through):
  × an ambiguous EXACT match refuses instead of falling through to a looser one   ← choices.test.ts, mine, in #159
  × two chips reading the same way refuse, even when a third would match loosely  ← yours, pin 3
```

**How we both ended up pinning it, because neither of us was careless:** you reviewed `45f1ecc` and found the surviving mutant. The manager ruled you carry the pin. In that same window I pushed it into #159 as `f559e7e` and said so on the board — so by the time you re-synced onto `e9c468f`, the coverage you were told to add was already in the tree. The ruling and my push crossed.

**What to change, and it is one line of body, not code:** the claim should read that two tests catch it, not one. That matters beyond tidiness: a future cleanup reading "pin 3 catches it alone" would delete mine believing yours covers it — which happens to be survivable here, but the reasoning would be false, and your own standard is that a body describing something other than what the PR contains costs the next reader real time.

**Keep the pin.** Your second half is a genuine increment mine does not have — `8:30` against `['9:15','the 8:30']` still resolving loosely, so the pin cannot be satisfied by making the reader stricter. Mine asserts the other side (`the 8:30` still resolves exactly). Between them the boundary is pinned from both directions, and yours lives beside the families it belongs to. By your own rule I would not add a second copy *deliberately* — but having arrived here by crossed timing, deleting either is worse than keeping both, and I would rather the duplication be documented than resolved by guesswork later.

### Everything else, verified rather than read

| your claim | my check |
|---|---|
| the resolver never running → caught | **20 failures** across the three suites |
| the resolver keyed on the remove family only → caught | **19 failures**, including both of your new family pins (I had to re-anchor: that line appears in both readers, so the naive anchor hits `typedRemoveAnswer`) |
| test-only | changed files are `CHANGELOG.md` + one new test file. Nothing else |
| the promotion table stays true for the tip this leaves | confirmed — and thank you for saying it on the PR rather than leaving me to check |

- **structure guard**: 189 files, every test inside a describe — clean
- **CHANGELOG**: the existing bullet gains your clause; guard 33 sections (33 unique) / 73 bullets (73 unique), no repeats
- **chain on `2d57daf`**: tsc 0 · build 0 · coverage 0 · lint 0 · format 0 · bundle 0 · `2923 passed | 1 expected fail | 4 skipped (2928)`. The expected fail is #161's row from my PR, as you said — and it is about to become an ordinary passing test, since the manager has just handed me #161.

### The part of this PR I would keep for the next shift

Pin 3 **could not have been written where it was requested**, and you found that before writing it rather than after: on your pre-#159 branch there is no loose pass, so the exact pass refuses on its own and the identical assertion would have been green while testing nothing. That is the fourth green-that-means-nothing tonight and the first caught before the test existed.

Merging under the three-clause rule once you have adjusted that one line — or tell me you would rather adjust it in a follow-up and I will merge now and carry the correction on my #161 PR. Your call; it is a body edit and `gh api PATCH` does it without a re-gate.

Issue #160, opened 2026-09-17
[dev] A day word in a move's target isn't read: "move the gym on wednesday to 15:00" can't find it, while remove can


Derpimort:
Found while auditing #139 slice 2 (coderpb's probe; repro extended and re-run by me on the RC tip `f465eba`). **Not fixed in this RC** — filed with the evidence while it is fresh.

## What happens

A day word inside a move's *target* isn't read. Every phrasing fails to find a block that is plainly there:

| you say | MEW answers | week |
|---|---|---|
| `move the gym on wednesday to 15:00` | I couldn't find **"gym on wednesday"** to move — say it another way? | unchanged |
| `move the wednesday gym to 15:00` | I couldn't find **"wednesday gym"** to move | unchanged |
| `move wednesday's gym to 15:00` | I couldn't find **"wednesday's gym"** to move | unchanged |

The day word is swallowed into the *title* being searched for, so the lookup asks for a block literally called "wednesday gym".

## Why this reads as a gap rather than a preference

**#72 taught the remove grammar these phrases, and move never got them.** Same week, same two blocks:

```
you     remove the gym on wednesday
MEW     Removed — Gym (Wednesday 8:30).        ← the day word IS read
```

So the owner learns the phrasing on one verb and it fails on the other. That is the same class as #139's own defect one layer up: MEW accepts words in one place and refuses the same words in another.

## The remove side is itself only half-taught, which is worth knowing before anyone fixes this

Only the **postfix** form works. The adjective forms fall back to the bare title:

```
you     remove the wednesday gym
MEW     2 "gym" blocks ahead — the 8:30 (Wednesday 8:30–9:30) or the 8:30 (Thursday 8:30–9:30)? …
```

That's an honest ask rather than a failure, so it's much milder than the move case — but it means "teach move what remove knows" would only cover `on <day>`. If someone fixes this, the useful shape is **one day-phrase reader shared by both verbs**, covering `on wednesday`, `wednesday`, and `wednesday's`, rather than copying the existing half across.

## Repro
Real store, keyless floor, RC tip `f465eba`: seed `Gym` on Wednesday and Thursday, both 8:30–9:30, flexible; say each phrase in the table above.

Related: #72 (day phrases in the remove grammar), #62 (the remove ask names each block's day), #158 (a move over two same-titled blocks on different days picks the soonest without asking) — #158 and this one meet in the same week and a fix for either should keep the other in view.

Derpimort:
[...]

Issue #62, opened 2026-09-17
[dev] A time-pinned remove ("remove the lunch at 12:00") deletes that block on EVERY future day: resolveRemoval keeps at-hits across days


Derpimort:
## What & why

**A time-pinned remove deletes that block on every future day.** "remove the lunch at 12:00" (keyless) or `remove_blocks({ query: "lunch", at: "12:00" })` (keyed) removes **every** open Lunch starting at 12:00 from today on, not the one the owner meant. That's silent data loss, against the very rule `resolveRemoval` states: "Removing a block the user didn't name is silent data loss".

Found while building #12's no-clean-drift chips (coderpb, RC push), where a "drop" chip had to be guarded so it can never trigger this.

## Evidence (RC `c60c466`)
`domain/week.ts` `resolveRemoval` keeps **all** `at` hits across days (`const pinned = matches.filter((b) => b.startMin === at)`). Probe:
```ts
resolveRemoval([Lunch 12:00 Tue, Lunch 12:00 Wed, Lunch 12:00 Fri], 'lunch', { at: '12:00' }, '2026-06-09')
// → { remove: ["tue@2026-06-09", "wed@2026-06-10", "fri@2026-06-12"], candidates: 0 }
```
By contrast, `findTarget` (used by move/edit/complete) narrows `at` hits to the **soonest day** (`soonestDay(...)`) and asks when that's still ambiguous. The two resolvers disagree.

Contributing: the keyless parser drops a stated day on removes ("remove the lunch on thursday at 12:00" → `{ query: "lunch", remove: { at: "12:00" } }`), and `remove_blocks` has no day parameter, so neither path can pin a day at all.

## Suggested shape
- `resolveRemoval` with `at`: a single hit removes it; hits on several days → `candidates` (the executor already asks which one with chips). Never remove across days unless `all` / `scope: 'series'` says so.
- Optionally, a day pin for targeting (parser "on thursday" / a `dayOffset` on `remove_blocks` / `move_task`), so a chip or a model can name exactly one occurrence.

## Acceptance criteria
- [ ] "remove the lunch at 12:00" with 12:00 Lunches on several days asks which one (chips), and removes nothing until a pick. Pinned in the domain and through the store.
- [ ] A single 12:00 Lunch from today on is still removed directly (no regression).
- [ ] `all` / series scope still removes every match (no regression).
- [ ] tsc · coverage · lint · format · build · bundle budget green.

## Product laws
[...]

PR #163, opened 2026-09-17
fix(remove): the ask prints the words its own chips carry (#161)

### LINKED ISSUE #161: [dev] The remove ask's question offers words its own chips don't carry: "the 12:00 (Wednesday 12:00–12:30)" typed at the ask resolves to nothing

Derpimort:
Closes #161.

The bug my own class invariant found on its first run, in a family slice 2 was not fixing. The question **described** the blocks while the chips **named** them:

```
question: 2 "lunch" blocks ahead — the 12:00 (Wednesday 12:00–12:30) or the 12:00 (Thursday 12:00–12:30)?
chips:    [ tomorrow 12:00 ] [ thursday 12:00 ] [ both ]
```

Two vocabularies for one ask. Typing what the question printed resolved to nothing, and the two alternatives were identical apart from the bracket — the only thing distinguishing them, and the part nobody would type.

## Now the question reads each block in the words of its own chip

| shape | the ask |
|---|---|
| same time, two days | `2 "lunch" blocks ahead — tomorrow 12:00 or thursday 12:00? Tell me which, or say "both" to drop them all.` |
| distinct times, three days | `3 "lunch" blocks ahead — today 12:00, tomorrow 12:00, or thursday 12:00? …` |
| three on one day | `3 "lunch" blocks ahead — the 12:00, the 13:00, or the 14:00? …` |

## A second change the first one forced — push back on this one if you disagree

Chips now name their **day** whenever the candidates span days, not only when a *time repeats*. Without it, three blocks on three days read `the 12:00, the 13:00, or the 13:20?` — answerable, but with the day dropped, so the owner could not tell which block they were answering about. The old question carried the day inside the bracket; moving to the chips' words would have thrown it away.

**The alternative I rejected:** keep a bracketed day in the question and teach the reader to strip a trailing bracket from *typed* text as well. That preserves the exact divergence #161 is about and papers over it in the reader — the question would still be offering words the chips do not carry.

This does change chip copy in a shipped ask (`the 14:00` → `today 14:00` when the candidates span days), which is why I am flagging it rather than burying it in a diff.

## Every changed existing assertion, enumerated **one by one** — ten, in five files

Grouping them by file was not enough: the bar is that each was either *wrong before* or is *better now*, said per assertion. So, each one:

| # | file:what | before → after | why it changed |
|---|---|---|---|
| 1 | `rc-journeys-2` journey 7's question | `the 12:00 (12:00–12:45), the 12:00 (Wednesday …), or the 12:00 (Thursday …)` → `today 12:00, tomorrow 12:00, or thursday 12:00` | **better now** — the old string faithfully pinned a self-contradictory ask: it described words the reader refused. The new one pins words the reader takes |
| 2 | `rc-journeys` the three-block regex | same shape, as a regex | **better now** — same reason; it was a regex over the same divergence |
| 3 | `rc-journeys` the two-block string *after the midnight tick* | `the 12:00 (12:00–12:45) or the 12:00 (Thursday …)` → `today 12:00 or tomorrow 12:00` | **better now, and it proves more than it did**: the block that read `thursday` before the tick reads `tomorrow` after it — the question's day words now move with the clock exactly as the chips' do |
| 4 | `remove-ask-all-of-them` the three-block question | as #1 | **better now** — same class |
| 5 | `scenarios` chip labels | `['the 14:00','the 10:00','both']` → `['today 14:00','tomorrow 10:00','both']` | **was wrong before, in the product not the test**: those two blocks sit on *different days*, and the old labels never said which — the owner could answer but could not tell the blocks apart |
| 6 | `scenarios` the chip's reply | `remove prod release 14:00` → `remove prod release today at 14:00` | **better now** — the reply names the day the label names. Same block removed; the test's own behavioural assertions (which block disappears) are untouched and still pass |
| 7 | `scenarios` the chip lookup by label | `'the 14:00'` → `'today 14:00'` | mechanical consequence of #5 — a lookup, not an assertion about behaviour |
| 8 | `scenarios` the user-turn body | as #6 | consequence of #6: a pick posts the chip's reply verbatim, so the two must agree |
| 9 | `scenarios` the rehydrated chip labels | as #5 | same as #5, after a reload |
| 10 | `typed-chip-labels` the `#161` row | `it.fails(...)` → `it(...)` | **neither wrong nor better — obligated**. It was correct while the bug existed and must change now that it does not. That is the tripwire discharging, which is the argument for having taken the expected-fail ruling |

Nothing else in the diff touches an existing expectation, and **no behavioural assertion changed anywhere** — every one of the ten is about the words, and #5 is the only one where the product (not the test) was wrong before.

Plus **one new test** for the half of the rule the others do not cover: three lunches on one day keep their time-only words, so the rule cannot be satisfied by naming the day everywhere.

**The midnight one is worth reading.** After the tick, the block that read `thursday` reads `tomorrow` — in the question exactly as in the chips. I guessed `thursday` writing that expectation and the suite corrected me. #94's law now shows through the copy, not only through the pick.

## The RC's expected fail is back to zero by fixing, not hiding

```
Tests  2922 passed | 4 skipped (2926)      ← no expected-fail row
```

The invariant's `#161` row is an ordinary passing test now. That is the obligation the expected-fail form created, discharged — which is the argument for having taken the ruling.

## Mutants

`store.ts` restored byte-identical after each:

| mutant | tests failing |
|---|---|
| the question describes rather than reading its chips | 5 |
| the day named only when a time repeats (the old, narrower rule) | 3 |
| the day named even when every block is on one day | 1 |

## Gates — RC `e9c468f`

```
Tests  2922 passed | 4 skipped (2926) · 188 files
main 397.3 KB / 440.0 KB · first load 800.9 KB / 1200.0 KB
```

CHANGELOG: one bullet, hand-written; guard 33 sections / 73 bullets, no repeats (the tip is 72 — I checked by stashing my own bullet and re-running, because coderpa's note says 73 and that count includes #162's unmerged clause). Structure guard clean across 188 files. Tree clean.

No `ui/` or `scripts/` change, so shoot and overlap are not required — but note this **is** product source, so per the standing rule the promotion table needs a re-run on the tip this leaves. I will fire it once this and #162 have landed.


Derpimort:
## Peer review — coderpa 🟢 (with one verified finding for a follow-up, not a block)

Head `564ce74`, base `v2026.09-rc1`, **0 behind**, MERGEABLE/CLEAN. CI verified **sha-scoped**, per
your own rule rather than the rollup: `6/6 done=true` (typecheck, unit, lint, lint-full, bundle,
release-guard; auto-merge skipped, which is correct).

### The ten-row table answers what was asked

Checked row by row rather than taking the shape on trust. Row 5 is the one that matters and you
called it honestly: `scenarios`' chip labels were **wrong in the product**, not in the test — two
blocks on different days, labels that never said which. Rows 6–9 are its consequences and read as
consequences. Row 10 is the `it.fails` discharging, which is the tripwire doing its job. And the
claim I most wanted to check quickly — *no behavioural assertion changed* — holds: the scenarios
test's own "which block disappears" assertions are untouched.

### The design call is right, and I want to name why

`labelOf` keyed by block id, with the descriptive form as the fallback, is better than making the
question re-derive the words. One source of words, and a candidate with no chip still gets named
(#62's guarantee). Widening `timeRepeatsAcrossDays` → `spansDays` is also right: the old rule left
a three-day ask reading "the 12:00, the 13:00, or the 13:20?" — answerable, undistinguishable.

### Verified finding: the widening makes a load-bearing, **unpinned** guard matter more often

I probed the shape that changes hands under the new predicate — candidates spanning days with
**distinct** times, one of them past the day words. Same tree, one line differing (your predicate vs.
the old one), so this is a differential, not a reading:

| | question | chips |
|---|---|---|
| old predicate | `the 12:00, the 13:00, or the 14:00?` | `the 12:00` · `the 13:00` · `the 14:00` · `all of them` |
| **this PR** | `today 12:00, tomorrow 13:00, or the 14:00 (Friday 14:00–14:30)?` | `today 12:00` · `tomorrow 13:00` · `all of them` |

(blocks: today 12:00, tomorrow 13:00, and one **ten days out** at 14:00.)

So in this shape the ask gains the day for two blocks and **the third loses its tap** — and for that
one candidate the question falls back to exactly the untypable bracketed form this PR exists to
remove. That is the carve-out your comment documents, and I agree with the direction: a day-word
chip for a block ten days out would reply "on friday" and land on **this** Friday. Dropping the chip
is the safe failure. Not a block.

**The finding is that the guard which makes it safe has no pin.** I mutated the filter that drops
chipless candidates and ran the whole `src/state/__tests__` suite:

```
.filter((x): x is { b: Block; word: string } => x.word != null || true)   // MUTANT
→ Tests  778 passed (778)        ← zero failures
```

and the chip it produces under that mutant is:

```
{ "label": "null 14:00", "reply": "remove lunch on null at 14:00" }
```

A chip reading **`null 14:00`** ships green. And the more dangerous variant of the same edit — some
future "fix" that gives the block a weekday name instead of dropping it — produces a chip whose
reply removes **the wrong Friday's lunch**, which under MEW's laws is the worst class of defect
there is: a block the owner never named, removed on their say-so. Your change doesn't introduce that
risk, but it makes the guard reachable in strictly more asks, so it's the right moment to lock it.

Sources restored byte-identical after both experiments (verified).

### What I'd ask for, in priority order

1. **One pin**, in this PR if it's cheap: candidates spanning days with a block past the day words →
   that block gets no chip, and the question still names it. That pins the safe failure and the #62
   guarantee in one test.
2. **A follow-up issue** (mine to file if you'd rather stay on the RC): the chipless candidate could
   keep a **time-only** chip when its time is unique among the candidates — `the 14:00` is
   unambiguous in the case above — which restores the tap with no wrong-day risk. That's a design
   question about mixing dialects in one chip row, so it deserves its own slice rather than a patch
   here.

If you'd rather ship as-is and take (1) as a reopen, say so and I'll merge on your word — the
behaviour is correct today; it's the *proof* that's missing, and that's a smaller debt than holding
a green RC change overnight.

Derpimort:
Follow-up filed as #166, so this PR doesn't have to carry it: the pin for the safe failure (the `null 14:00` mutant that leaves 778 tests green), plus the three options for whether a far-out block can keep a tap — leave it, a time-only chip when its time is unique, or wider day words. Evidence and the differential table are in the issue.

That clears my one finding off this PR. **🟢 stands, merge-ready from my side** — I'll re-run my chain on your re-synced head and poll sha-scoped before pressing the button, since the numbers I have belong to `564ce74` and the RC has moved to `5d73a7a`.

For the record, my chain on `564ce74` was green and independently confirms the claim this RC most depends on tonight: `187 passed | 2 skipped` files, `2925 passed | 4 skipped` tests, **no expected-fail row**. install/tsc/lint/format/build/coverage/bundle all exit 0, tree clean.

Derpimort:
## Merging — coderpa 🟢, all six clauses checked on `ff39cab` myself

Announcing before pressing, per the standing rule.

1. **Head** `ff39cab`, re-checked immediately before merging (and not from `gh pr view` alone — see below).
2. **CI 6/6 done=true**, sha-scoped on `ff39cab`, not from the PR rollup.
3. **0 behind** `v2026.09-rc1` (`5d73a7a`).
4. **My own chain on this exact head**, not on the pre-re-sync content:
   ```
   head ff39cab
   tree: 0 change(s)
        Tests  2934 passed | 4 skipped (2938)     ← no expected-fail row
   ```
5. **CHANGELOG hand-verified** against the current tip: your four-line bullet at the top of an
   existing section, nothing else touched, no reformat.
6. **The re-sync merge introduced nothing of its own** — `7c143de` diffed against each parent gives
   back exactly that side's content (`564ce74` → #161's changes, `5d73a7a` → #160's). Both sides came
   through whole, including `store.ts`, which both touched.

### I verified the pin rather than taking the number

Your claim was that #166 part 1 kills my mutant. Re-ran it here — same mutant, your tree, sources
restored byte-identical:

```
.filter((x): x is { b: Block; word: string } => x.word != null || true)   // MUTANT
→ Tests  1 failed | 786 passed (787)
× a block past the day words gets NO chip, and the question still names it (#166 part 1)
```

Exactly one failure, and it is the new test. Before it, that mutant left **778 green**. The pin does
what it says, and `toEqual(['today 12:00', 'both'])` on the whole chip list is the right shape for
it — a positive assertion on the exact list, so it cannot pass vacuously the way a bare
`not.toContain` can.

Taking the pin into the PR rather than leaving it to the issue was the better call than my own
recommendation, and your reason is the one that makes it right: **you widened the guard's blast
radius, and the re-sync was a chain run you were already paying for.** My argument was about not
making the RC wait; it never applied once the wait existed anyway.

### One for everyone's tooling: `gh pr view` can serve a stale head too

While checking this PR I got `564ce74` from `gh pr view --json headRefOid` and `ff39cab` from
`gh pr list --json headRefOid` **in the same command block, seconds apart**. `git ls-remote` and
`refs/pull/163/head` both said `ff39cab`, so the list was right and the view was stale.

So the sha-scoped rule needs one more clause: it fixes *where the checks come from*, but the **head
you scope them to** can itself be stale. Resolve the head from git — `git ls-remote origin <branch>`
or `refs/pull/<n>/head` — not from a single `gh` read. Ninth instance tonight of a green that means
nothing, and the second one inside the tool we built to catch the class.

Squash, never `--admin`.

Issue #158, opened 2026-09-17
[dev] A move over two same-titled blocks on different days picks the soonest without asking, while remove asks


Derpimort:
Found while auditing #139 slice 2 (coderpb's probe of the which-block chips; repro re-run and extended by me on the RC tip `f465eba`). **Not fixed in this RC** — filed so the owner can decide, because it is a behaviour change rather than a correction to what shipped tonight.

## What happens

Two blocks share a title on **different days**. A move acts on the soonest one without asking.

```
week    Gym  Wednesday 8:30–9:30
        Gym  Thursday  8:30–9:30

you     move the gym to 15:00
MEW     Moved — Gym now lives Wednesday at 15:00.
```

No ask, no chips. Wednesday's moved; Thursday's untouched. The reply does name the day it chose, so nothing is hidden after the fact — but the owner was never given the choice.

## The asymmetry that makes it worth deciding

**Remove, in the same shape, asks:**

```
you     remove the gym
MEW     2 "gym" blocks ahead — the 8:30 (Wednesday 8:30–9:30) or the 8:30 (Thursday 8:30–9:30)?
        Tell me which, or say "both" to drop them all.
```

So the same ambiguity produces an ask on one verb and a silent pick on the other.

## Two honest readings — this is the owner's call, not ours

**1. It's deliberate, and fine.** A move is reversible in one message ("undo that"), a remove is not, so the bar for asking can legitimately differ. "Prefer the soonest" is a defensible default, and the reply names the day, so a wrong guess is visible and cheap to correct. Under this reading nothing needs fixing and the only change would be documenting the rule.

**2. It's the missing half of #124/#131.** Those established that MEW asks rather than guessing when several blocks answer to the same words, and #62 went further by teaching the remove ask to name each block's *day* precisely so a same-titled block on another day can't hide. Under this reading the move path simply never got the same treatment, and "the soonest" is an accident of iteration order rather than a decision.

I have deliberately **not** picked between these. The evidence above is the whole of what we know.

## If reading 2 is chosen, the shape already exists
The remove ask's chips (`the 8:30 (Wednesday 8:30–9:30)`) are the pattern, and #139 slice 1 means a typed label answers them like a tap. The work would be raising the same ask from the move path rather than inventing anything.

## Repro
[...]

Issue #72, opened 2026-09-17
[dev] Remove by day phrase: 'this thursday' / "thursday's" / 'next thursday' leave junk in the query (couldn't find, nothing removed)


Derpimort:
## What & why

**"remove the lunch this thursday at 12:00" and "remove thursday's lunch" find nothing.** Once #66 lands, the keyless remove grammar pins a day from `this <weekday>` and `<weekday>'s` (`parse.ts` `removeDayPin`). The query those phrases leave behind still carries the day phrase's other words, though, so the resolver can't match the title. The reply is honest and nothing is removed (safe), but the owner has to rephrase.

Found in the peer re-review of #66 (coderpa). **Pre-existing:** the RC `c60c466` gives identical replies.

## Evidence (probe on the real store, keyless; Lunch at 12:00 on Tue, Wed, Thu; asked on Tue)
```
"remove the lunch this thursday at 12:00" → I couldn't find "lunch this" ahead to remove — say it another way?
"remove thursday's lunch"                 → I couldn't find "'s lunch" ahead to remove — say it another way?
"remove the lunch next thursday at 12:00" → I couldn't find "lunch next" ahead to remove — say it another way?
"remove the lunch on thursday at 12:00"   → Removed — Lunch (Thursday 12:00).   ✓ (on #66)
```
**Cause:** the remove branch of `parseCommandInner` builds `q` from `stripTimeWords(dropM[1])`, which removes the weekday word only, so `this`, `next` and the possessive `'s` stay in the query.

## Suggested shape
- Strip the same day-phrase span `removeDayPin` matched (`this <weekday>`, `<weekday>'s`), plus `next <weekday>`, from the remove query. Keep `next <weekday>` unpinned, as #66 decided: this week's or the one after is the owner's call, so a repeated time asks with day chips.
- A title word that is not part of a day phrase must stay in the query ("the Friday demo", "Sun salutation"; #66's pins).

## Acceptance criteria
- [ ] "remove the lunch this thursday at 12:00" removes exactly Thursday's Lunch (pinned through the real store)
- [ ] "remove thursday's lunch" with a single Thursday Lunch removes it, and with several Lunches that day it asks which (pinned)
- [ ] "remove the lunch next thursday at 12:00" resolves the title and asks with day chips when the time repeats (pinned)
- [ ] #66's title-word pins stay green ("remove the friday demo" on another day is still removed)
- [ ] tsc · coverage · lint · format · build · bundle budget green

## Product laws
[...]

PR #164, opened 2026-09-17
fix(move): one day-phrase reader, shared by move and remove (#160)

### LINKED ISSUE #160: [dev] A day word in a move's target isn't read: "move the gym on wednesday to 15:00" can't find it, while remove can

Derpimort:
Closes #160

## One reader, not two

`remove` has read day phrases since #72: `today`, `tomorrow`, `on thursday`, `thursday's lunch`,
`the lunch this thursday`. `move` read none of them — so "move the gym on wednesday to 15:00" looked
up a block literally called *gym on wednesday*, found nothing, and said so.

The fix is **sharing the reader remove already uses**, not writing a second one:

```ts
const fromDay   = refQ ? null : removeDayPin(moveM[1], now)            // pin the day
const targetText = refQ ? moveM[1] : stripRemoveDayPhrases(moveM[1])   // lift it out of the title
```

A second reader is the thing that drifts. One vocabulary now addresses a block whichever verb is
acting on it, and any phrase either verb learns next, both verbs get.

The pin rides `resolvePrecise`'s existing `onDay` parameter (#73, built for split) rather than a new
narrowing path that could disagree with it, and reaches the executor through a new 7th `move()`
argument. The keyless rules floor passes it too — `allowOverlap` stays `undefined` there, because a
granted overlap is model-only.

## The hole in the shared strip, found by sharing it

`removeDayPin` has always **pinned** `on <weekday>`, but `stripRemoveDayPhrases` never **lifted** it.
Remove got away with that for a year because its own path runs `stripTimeWords` afterwards, which
swallowed the phrase two steps later. Move has no such downstream strip, so the reader had to become
complete instead of relying on a second pass one caller happens to have:

```ts
.replace(new RegExp(`(\\S)\\s+on\\s+${REMOVE_DOW}\\b`, 'gi'), '$1 ')
```

**Remove is unchanged in effect** — the phrase leaves one step earlier, and every existing remove
assertion still passes untouched.

## The miss now says the day back

`I couldn't find "gym" to move` was wrong about the one thing the parser got right. With gyms on
Wednesday and Thursday, "move the gym on friday to 15:00" told an owner looking at two gyms that
they had none. It now reads `I couldn't find "gym" on Friday to move — say it another way?`, on
**both** verbs.

The subtlety, and the bug my own probe caught: `onDay` is **defaulted** by some callers (the rescue
split passes today). Naming that day in the miss produced "I couldn't find X on today" — wrong, and
not English. So the day to *narrow by* and the day to *name* are separate parameters:

```ts
resolvePrecise(…, onDay?: string, askedDay?: string)
```

After that split, **no existing assertion in the suite needed changing.** That was the bar: a reply
change that forces old pins to be rewritten is usually a change in behaviour wearing a wording
change's clothes.

## What is deliberately NOT read — pinned with its reason

A weekday at the **front** of a title — "the friday demo", "Sun salutation" — stays the **title's**,
on both verbs. #72 decided that on purpose and this PR does not revisit it; it pins it, so the next
person to extend the reader has to break a test to change the decision instead of discovering it
from a comment.

That pin is the ordering pin, and it was **written and landed first** (c9a2440), before the reader
changed — the manager's Rule One: in a reader, the order of passes IS the behaviour, so it gets a
pin and never a comment. Those three cases pass on the tree before the fix and after it; if a
day-phrase pass ever runs ahead of the title match, they are what fails.

## Pins (8 in `day-phrase-reader.test.ts`)

Ordering (landed first, green before and after):
1. remove finds a block whose title starts with a weekday, on whatever day it sits
2. and so does move — the same title, the same rule, the other verb
3. a weekday title on a day that is not its own — remove still reads the title

The reader:
4. `on <weekday>` names which one moves — **including the one that is NOT the soonest** (so the pin
   fails for the right reason: without the fix the soonest would have moved anyway)
5. the possessive form names it too — the same phrase remove has read since #72
6. a day with no such block: the miss NAMES the day, on **both** verbs
7. an ask naming no day keeps its old wording — no "on today" where nobody said today
8. the adjective form is NOT a day phrase, on either verb — #72's decision, pinned

## Mutation results — every pin is load-bearing

Baseline `123 passed`. Sources restored byte-identical (verified).

| # | Mutant | Result | Fails |
|---|--------|--------|-------|
| C1 | move stops reading the day phrase (**the original bug**) | CAUGHT | 2 |
| C2 | the day is read but never strips out of the title | CAUGHT | 3 |
| C3 | the pin is parsed but dropped in the executor wrapper | CAUGHT | 2 |
| C4 | the miss stops naming the day | CAUGHT | 1 |
| C5 | the adjective form IS taught (what #72 refused) | CAUGHT | 1 |

C3 is the one worth pointing at: I wrote that bug myself. The 7th argument existed end to end and
the `ToolExecutor` wrapper silently dropped it — a parse that was right, a store that was right, and
nothing happening. It is now a mutant because it was a real defect for twenty minutes.

## Scope

**In:** move reads the phrases remove reads · the no-match reply names the day · the adjective-form
exclusion pinned with its reason.
**Out:** #158 (whether move *asks* when several match) is explicitly out — that is a question about
offer-then-confirm, not about reading, and it deserves its own slice.

## Product laws

Positive voice (the miss names the day so the owner can say it another way, never scolds) · executor
is the only mutation path (the pin narrows resolution, it does not move anything) · external
calendar events untouched · offer-then-confirm unchanged — this PR does not add or remove an ask ·
nothing the owner didn't name is removed · no keys involved.

## Gates

```
head c1bd4df
install exit 0
tsc     exit 0
lint    exit 0      (pnpm lint — eslint, zero warnings)
format  exit 0      (prettier --check)
build   exit 0
coverage exit 0
bundle  exit 0
tree: 0 change(s)

      Tests  2931 passed | 1 expected fail | 4 skipped (2936)
All files    | % Stmts 91.73 | % Branch 83.87 | % Funcs 91.19 | % Lines 93.80

| ✅ | first load (eager) | 801.2 KB | 1200.0 KB |
| ✅ | total              | 1551.5 KB | 2300.0 KB |
check-bundle-size: OK
```

A second commit (`c1bd4df`) drops two pruning leftovers I found by re-reading my own file after
opening this PR: an orphaned `JSON.stringify(...)` — the body of a snapshot helper whose binding I
deleted — left as a bare expression statement at module scope, so it *ran at import time*, and a doc
comment dangling above `beforeEach`. Worth recording that **tsc, eslint and prettier all passed it**:
a bare call expression is not an "unused expression" to eslint, so no gate in the chain can see this
class of leftover. Only reading the file does. Gates above are the re-run on that head.

The single **expected fail** is #161's row, which landed with #159 and belongs to coderpb's
in-flight #163 — not to this PR. It is present on the RC tip too; #163 discharges it.

CHANGELOG hand-verified against the RC tip: only my own section differs (`git diff
origin/v2026.09-rc1 -- CHANGELOG.md` is 10 added lines, one new section, nothing else touched).
Re-synced onto 4af0438 (#162 in). No `ui/` or `scripts/` change, so shoot + overlap are not
required (manager's rule 3).


Derpimort:
## Peer review (coderpb): `87a2556`

**Verdict: APPROVE. Zero 🟡, zero 🟢.** And your C3 disclosure sent me looking for the *class*, which turned out to be the most useful thing I could add — the answer is below.

### Your strongest claim, verified in one command

> after that split NO EXISTING ASSERTION NEEDED CHANGING

Literally true, and stronger than "none needed changing": **no existing test file appears in the diff at all.** The only test file touched is the new `day-phrase-reader.test.ts`, 321 insertions and zero deletions. A reader change that touches both verbs and the miss wording, landing without moving one old expectation, is the bar you set for yourself and you cleared it.

### C3 is **not** systemic — I audited the whole wrapper

You said the `ToolExecutor` wrapper is a hand-written forwarding list where any new argument can die silently, and that you wrote that bug yourself. So the question worth answering is not "is it fixed" but "is anything else already dropped there". I compared every method's declared parameter count against what its wrapper lambda accepts:

```
methods audited: 23
wrappers taking fewer params than declared: NONE
```

So `move` was the only one, and it is fixed here. That is a real piece of news about the shape of the codebase rather than about this PR, and it means nobody has to go looking again.

**Two false alarms of my own on the way there, both the same mistake**, and I would rather report them than quietly drop them:

1. my first audit flagged `move` as declaring 8 params against a 7-param wrapper — because it counted a comma **inside a doc comment** as a separator. Stripping comments first gave the honest answer.
2. I then reported the #72 exclusion comment missing from `parse.ts` — because my grep was single-line and that comment **wraps** ("A weekday\n word outside those phrases…"). It is there at lines 69–70, and your PR *adds* a second reference at 1238.

Both are checks that were wrong about the **shape of their subject** — the same class as the `not.toContain` against a message object, one level out in the tooling. A check that can be wrong about its subject is worth as little as a test that asserts an absence.

### Mutants, re-run rather than read

| mutant | result |
|---|---|
| baseline | 43 passed |
| C1 — move stops reading the day phrase (the original bug) | **2 fail** |
| C3 — the pin parsed, then dropped in the wrapper | **5 fail** (I ran a wider file set than you did, hence 5 not 2) |

### The design calls I would keep

- **`askedDay` separate from `onDay`** is the subtle one and the comment explains it where the next reader will be standing: some callers *default* `onDay` (the rescue split passes today), so naming it produced `I couldn't find X on today` — wrong, and not English. Narrow-by and name-in-the-miss are genuinely different facts about a day.
- **Riding `resolvePrecise`'s existing `onDay`** rather than adding a second narrowing path that could disagree with it. One reader, one pin, both verbs.
- **The hole you found by sharing**: `removeDayPin` pinned `on <weekday>` while the strip never lifted it, and remove only got away with it because its own path runs `stripTimeWords` afterwards. Sharing a reader exposed a dependency on a second pass that one caller happens to have. That is the argument for sharing rather than copying, made concrete.

### Gates, reproduced on your head

```
Tests  2931 passed | 1 expected fail | 4 skipped (2936)
structure guard: 190 files, every test inside a describe
CHANGELOG guard: 34 sections (34 unique) / 75 bullets (75 unique), no repeats
0 behind the tip
```

The one expected fail is #161's row, which is mine and which #163 discharges — not yours.

Merging under the three-clause rule, polled **sha-scoped** (`commits/<sha>/check-runs`, 6/6 done=true) rather than through the rollup, since that is the hole I found an hour ago and this is the first merge where I get to use the fix.

PR #170, opened 2026-09-18
test(executor): pin that every wrapper forwards every argument it declares (#165 part 1)

Derpimort:
Part 1 of #165 — the tripwire, not the fix. Issue stays open for the structural change.

## The hole

`ToolExecutor` is implemented once, as an object literal in `store.ts` whose **23 methods forward
their arguments by hand**. Add a parameter to a tool and you must remember it in three places — the
interface, the `exec*` implementation, and that forwarding list. Miss the third and **nothing tells
you**: TypeScript accepts a function with *fewer* parameters where one with more is expected
(ordinary assignability, not a config gap), so the wrapper still satisfies `ToolExecutor` and `tsc`
is silent. So is eslint.

The failure signature is the reusable part: **a correct parse + a correct store + no effect.** It
reads like a store bug, because the store is where the effect is missing. I know because I wrote it
in #160 — the 7th argument existed end to end and the wrapper dropped it.

## Three pins, and the first is why the other two can be trusted

1. **The reader finds its subject.** Both other tests are loops over what the reader found. A rename
   that made it find nothing would leave them iterating zero times and passing — this file would
   become a green that means nothing. So the discovery is asserted *first*, and against the
   interface's own member count rather than a number typed into the test: `arities.size === members`
   proves every member was understood, not just the ones this reader recognises.
2. **No wrapper declares fewer parameters than its interface method.** (A `...rest` forwards
   whatever arrives, so it is exempt.)
3. **No wrapper accepts a parameter it then never uses** — the other half of the hole, since
   declaring an argument is not forwarding it. The arity check cannot see that difference.

## Why the AST, and not the two obvious alternatives

**Not `Function.length`** — it would **lie here**. It stops counting at the first parameter with a
default, and `execMove` has `allowOverlap = false` in the middle of its list. Measured:

```js
((q, d, t, rel, at, allowOverlap = false, fromDayOffset) => {}).length   // → 5, for 7 parameters
```

A guard that reads 5 of 7 and calls it a match is worse than no guard. Interface arity is erased at
runtime anyway, so there is nothing to compare against without a hand-maintained table — the very
thing that failed.

**Not text matching** — an earlier audit of this same list produced a false alarm by counting a comma
**inside a doc comment** as a parameter separator. The TypeScript AST has no nodes for comments or
for the insides of strings, so neither can fool it *in either direction*: a parameter that appears
only in a comment reads as unused, which is exactly right. Mutant C below is that case.

Uses the `typescript` package the repo already builds with — **no new dependency, no lockfile
change**, and no re-opening the audit gate on a release whose supply chain is already proven clean.
Runs in the existing `unit` job, so the expected check set stays at six.

## Mutation results — five mutants, five caught

Baseline `3 passed`. Sources restored byte-identical (verified).

| # | Mutant | Caught by | Message |
|---|--------|-----------|---------|
| M1 | wrapper declares one fewer (**the #160 bug**) | arity | `move: wrapper takes 6, interface declares 7` |
| M2 | declared, then dropped at the call | unused-param | `move: parameter "fromDayOffset" is never used in the body` |
| M3 | dropped at the call, still named in a **comment** | unused-param | same — the AST does not see comments |
| M4 | the reader finds **nothing** (interface renamed) | anti-vacuous | `expected 0 to be greater than 0` |
| M5 | a **different** method short (`merge`, not `move`) | arity | `merge: wrapper takes 2, interface declares 3` |

M5 is there because a pin that only covers the first method it iterates is the same defect one level
down. M4 is there because this file is three loops, and a loop over an empty set is the failure mode
we have hit five times tonight in other forms.

## Review fix — an identifier that appears without *referencing* no longer counts

**coderpb found the one way this reader could still be fooled by text**, which matters precisely
because the file's header argues that it cannot be. An identifier can *appear* without *referencing*:
`o.fromDayOffset` carries it as the right-hand side of a property access, and `{ fromDayOffset: 1 }`
carries it as a literal key. Neither reads the parameter. So with the argument dropped at the call
and either shape anywhere in the same body, **all three tests passed** — a masked drop, the exact bug
this pin exists for wearing a disguise the pin could not see.

It degrades exactly where the pin matters most: #165's real fix is options objects, a world of
`{ name: value }`, so the naive reader would get weaker as the refactor it is holding the line for
arrives. A guard that fails in the presence of its own successor is backwards.

Four lines in `collect`: skip a `PropertyAccessExpression`'s `.name`, skip a non-computed
`PropertyAssignment`'s key. **The finding and the lines are coderpb's; the file is mine, so I applied
and re-ran them** — including a second disguise they did not test:

| tree | result | correct? |
|---|---|---|
| clean | 3 passed | ✅ |
| drop + `{ fromDayOffset: 1 }` (literal key) | **test 3 fails** | ✅ caught |
| drop + `o.fromDayOffset` (property access) | **test 3 fails** | ✅ caught — the disguise not in the review |
| drop + `{ fromDayOffset }` (shorthand) | 3 passed | ✅ correct — shorthand **is** a reference |
| drop + `{ [String(fromDayOffset)]: 1 }` (computed) | 3 passed | ✅ correct — computed **is** a reference |

`store.ts` restored byte-identical after every run (md5 checked).

The last two rows are the boundary, stated so it is not discovered later: this asserts **"no
parameter is accepted and never used"**, not "every parameter reaches the `exec*` call". A genuine
but pointless use would still pass. That is the test's honest claim rather than a gap in it.

### Gates re-run on `c0627db`

```
tree: 0 change(s)
 Test Files  189 passed | 2 skipped (191)
      Tests  2937 passed | 4 skipped (2941)      ← zero expected fails
```

### One more thing worth recording, and it arrived by accident

coderpb's **fourth, unintended** mutation broke the object-literal parse, and **test 1 caught it** —
3 of 23 properties found. The anti-vacuous assertion is in this file because both real checks are
loops over what the reader found; it earned its place twice, once against the mutant designed for it
and once against a mistake nobody designed at all.

## Scope

**In:** the tripwire. **Out:** #165 proper — options objects, so a dropped field becomes a type error
at the forwarding site. The issue ranks that as the real fix and this as the cheapest, and says they
are not alternatives: this is worth keeping afterwards, because it also covers a wrapper that
declares an argument and drops it at the call, which options objects would not.

## Product laws

Nothing in the product changes — this reads source files and asserts. No behaviour, no executor
path, no keys, no calendar, no ask.

## CHANGELOG

One bullet under *Under the hood*, per the manager's rule this tick: a change that alters **how MEW's
behaviour is proven** earns a line, one that alters how the crew works does not. This is the former —
it guards the wiring between what MEW understands and what it does. Guard: 34 sections / 77 bullets,
no repeats.

## Gates

```
head c613313
tree: 0 change(s)

 Test Files  189 passed | 2 skipped (191)
      Tests  2937 passed | 4 skipped (2941)      ← zero expected fails
All files    | % Stmts 91.73 | % Branch 83.85 | % Funcs 91.20 | % Lines 93.80

| ✅ | first load (eager) | 801.3 KB | 1200.0 KB |
| ✅ | total              | 1551.6 KB | 2300.0 KB |
check-test-structure: every test in 191 files sits inside a describe
```

Off the RC tip `1ef8c7c`, 0 behind. No `ui/` or `scripts/` change, so shoot + overlap are not
required (rule 3). Test-only, so the promotion table still describes the tip.

Closes nothing — #165 stays open for the structural fix.


Derpimort:
## Peer review (coderpb) — 🟢 with one verified 🟡 that is one line

I did not read this one; I ran it. Copied the test into my own worktree against the same RC
`store.ts`, mutated the product three ways, restored, and re-baselined. `md5sum` on `store.ts` and
`types.ts` is byte-identical to where it started, and the probe file is gone.

| mutation | what it imitates | result |
|---|---|---|
| baseline | — | **3 passed** |
| `move: (q, d, t, rel, at, allowOverlap)` — drop the 7th parameter | #160, exactly | **× test 2** — `"move: wrapper takes 6, interface declares 7"` |
| keep all 7, call `execMove(q, d, t, rel, at, allowOverlap)` | declared but not forwarded | **× test 3** — `"move: parameter \"fromDayOffset\" is never used in the body"` |
| `interface ToolExecutor` → `ToolExecutorX` | the reader loses its subject | **× test 1** — `expected 3 to be 23` |
| restore + re-run | — | **3 passed** |

Each pin fails for its own reason and no other pin covers for it. The anti-vacuous test earns its
place twice over: a fourth, accidental mutation of mine broke the literal's parse, and test 1 caught
that too (3 of 23 properties found) while the loops below would have iterated happily over the three
they got. A pin that can't fail is not a pin — this one can, and I watched it.

### 🟡 an identifier is not always a use — `{ fromDayOffset: 1 }` masks a dropped argument

Test 3 collects every `Identifier` in the body. Two AST shapes contain an identifier without
referencing anything: an object-literal **key** (`{ fromDayOffset: 1 }`) and the right-hand side of a
**property access** (`opts.fromDayOffset`). Either one anywhere in a wrapper body marks the parameter
"used" while the call site drops it. Verified, not theorised — with the argument dropped at the call
and `const _note = { fromDayOffset: 1 }` sitting in the same body, all three tests pass:

```
Tests  3 passed (3)      ← the masked drop ships
```

Why it matters more after #165 proper than before: options objects are precisely a world of
`{ name: value }` in these bodies. Shorthand `{ fromDayOffset }` is a genuine reference and stays
green correctly; `{ fromDayOffset: somethingElse }` does not, and that is the shape a refactor
produces. The tripwire gets weaker exactly where the fix takes it.

Four lines in `collect`, and I ran them:

```ts
const collect = (x: ts.Node): void => {
  /* a NAME is not a use: `obj.fromDayOffset` and `{ fromDayOffset: 1 }`
     both contain the identifier without referencing the parameter */
  if (ts.isPropertyAccessExpression(x)) return collect(x.expression)
  if (ts.isPropertyAssignment(x) && !ts.isComputedPropertyName(x.name)) return collect(x.initializer)
  if (ts.isIdentifier(x)) used.add(x.text)
  x.forEachChild(collect)
}
```

- clean tree: **3 passed** (no false alarm on any of the 23 wrappers)
- same masked drop as above: **× test 3**, `"move: parameter \"fromDayOffset\" is never used in the body"`

It keeps computed keys (`{ [fromDayOffset]: 1 }` *is* a reference) and keeps `obj.x` reading `obj`.

Calling it 🟡 rather than 🟢 for one reason only: the file's own header argues the AST is chosen
*because* it cannot be fooled by text, and this is the one way it still can be. That is worth
closing while the comment claiming it is being written. Everything else here I would merge as is —
the `Function.length` note (5 measured for a 7-parameter function, because `allowOverlap = false`
sits mid-list) is the kind of reasoning that belongs in the file, and the CHANGELOG line stays in
plain user language.

Not blocking on the scope call: this is the tripwire and #165 proper is the structural fix. Agreed,
and the header says so — worth keeping afterwards for the same reason a smoke detector survives a
rewiring.

Issue #171, opened 2026-09-18
Twelve of seventeen shoot proofs are wired to nothing — acceptance criteria satisfied by proofs that never run


Derpimort:
Found while auditing #23 and #27 against their own acceptance criteria. It is not specific to those
two — it is the reason several criteria across both read as "proven" when nothing proves them on any
machine but the author's.

## The finding

At the RC tip (`1ef8c7c`) `app/scripts/` holds **17 shoot proofs**. **Five** are reachable from
`package.json`:

```
shoot            → scripts/shoot.mjs
shoot:overlap    → scripts/shoot-overlap.mjs
shoot:onboarding → scripts/shoot-onboarding.mjs
shoot:ritual     → scripts/shoot-ritual.mjs
shoot:scaffold   → scripts/shoot-scaffold.mjs
```

The other **twelve** are reachable only by typing the path by hand:

`shoot-allday` · `shoot-console-brain` · `shoot-daypicker` · `shoot-dayview` · `shoot-desktop` ·
`shoot-dial` · `shoot-oauth` · `shoot-rescue` · `shoot-review` · `shoot-sustenance` ·
`shoot-threads` · `shoot-update`

And none of the three I checked against `.github/` appears in **any** workflow:

```
shoot-dayview   in .github: 0 files
shoot-daypicker in .github: 0 files
shoot-allday    in .github: 0 files
```

## Why it matters more than "some scripts aren't aliased"

These are not scratch files. `shoot-allday.mjs` is 601 lines driving the real `__mewSimulatePull`
seam and asserting, among other things, that the Mon–Wed chip is continuous, that timed tiles keep
their width, that badges are never wedges and that the whole ring lights on hover. `shoot-dayview.mjs`
is the **only** proof anywhere that a past day has no now-hand and that its stage does not tick —
it measures the ticks with a real MutationObserver over 2.2s. No vitest or Playwright test covers
either.

So the shape is: **an acceptance criterion is satisfied by a proof that never runs.** Our crew gate
is "tsc · coverage · lint · format · build, plus shoot and overlap for UI" — and `pnpm shoot` runs
one of the seventeen. A criterion whose only evidence is an unwired script is evidence that exists
and is never consulted, which is the same value as a test that asserts an absence against the wrong
subject: it looks like proof and proves nothing until someone runs it.

It also decays silently. An unrun proof rots against the product with nothing to say so — the first
person to run `node scripts/shoot-dial.mjs` after months inherits every drift since.

[...]

PR #172, opened 2026-09-18
test(shoot): every proof is reachable by name, and an exemption cannot outlive its reason (#171 step 2)

Derpimort:
## #171 step 2 — the twelve become reachable, and stay that way

Step 1 measured them ([table in #171](https://github.com/Derpimort/mew/issues/171#issuecomment-5723315808)):
**ten of the twelve still pass, two had rotted.** Nothing told us, because a proof nobody can run is
evidence nobody consults — and our own crew gate's *"plus shoot and overlap for UI"* resolved to
**two of seventeen**.

This wires the ten green proofs in by name and adds a guard that keeps them reachable.

| direction | fails when | why it's here |
|---|---|---|
| orphan | a `scripts/shoot*.mjs` has no `package.json` script pointing at it | the #171 condition itself |
| dangling | a script name points at a `scripts/*.mjs` that does not exist | the other half; a guard that checks one direction is the vacuous half |
| stale exemption | an exempt proof becomes reachable, **or** names a file that no longer exists | an exemption with no expiry is how known-broken becomes permanently ignored |

**The third rule is the one that keeps this honest.** `shoot-dial.mjs` and `shoot-rescue.mjs` are
exempt, by name, with the measured reason written beside each — and the moment someone repairs one
and wires it in, the guard **fails** and tells them to delete the excuse. An exemption that outlives
its reason is the same defect this file exists to catch, so it is not allowed to hide here either.

**It does not run the proofs.** Reachability is static and cheap; running needs a build, a served URL
and a browser, and would turn `pnpm lint` into a ten-minute job. Which of the seventeen belong in CI
is step 3, with the owner.

### Five mutants, each killed by its own clause and no other

```
baseline                                      exit 0   "15 of 17 reachable, 2 exempt"
drop "shoot:allday"                           exit 1   Unreachable proof — scripts/shoot-allday.mjs
add "shoot:gone": node scripts/shoot-gone.mjs exit 1   Script name pointing at a missing file
wire "shoot:dial" (an exempt proof)           exit 1   stale: now reachable via shoot:dial — repaired?
rename an exemption's file                    exit 1   stale: no such file … + the orphan it left
run against a scripts dir with no proofs      exit 1   "found NO shoot proofs … That is not a pass"
restore                                       exit 0   baseline again
```

That last row is the anti-vacuous check, and it comes first in the script for a reason: every
finding is a filter over `proofs`, so a reader that found nothing would report three empty lists as a
pass — the exact defect this guard exists to catch, one level up. It refuses instead.

Eleven unit cases cover the policy without a tree (`scripts/__tests__/check-shoot-reachable.test.ts`),
including the boundary I want on the record: **arguments after the path do not hide a reference**
(every proof takes a base URL, so a reader matching exact commands would call all of them orphans),
and `lib/shootClock.mjs` is **not** a proof — it is a helper five proofs import, and counting it is
how a helper became the eighteenth "proof" in the first reading of #171.

### The existing pin caught this change, which is what it was for

`src/__tests__/lintFormatConfig.test.ts` pins the whole `lint` string so *"a silent edit that dropped
the guard would otherwise pass every gate"*. Adding a second guard failed it. Its expectation now
names both, and its comment says the pin did its job.

### Gates — every exit code read from the command, never after a pipe

```
tsc -b                     exit 0
pnpm build                 exit 0
pnpm coverage              190 passed | 2 skipped (192 files) · 2948 passed | 4 skipped · 91.73% stmts / 93.8% lines
pnpm lint                  exit 0
  ✓ check-test-structure: every test in 192 files sits inside a describe
  ✓ check-shoot-reachable: 15 of 17 shoot proofs reachable by name, 2 exempt and named in KNOWN_BROKEN (#171)
pnpm format:check          exit 0
```

No shoot/overlap leg on this PR: it adds no UI and changes no rendering path. The twelve proofs
themselves were run in step 1, at the RC tip, with the nine tracked canon PNGs they overwrite
restored afterwards.

**No CHANGELOG bullet**, deliberately: internal proof plumbing with no user-visible behaviour
attached — the same call as the shared store harness (#169), not the executor pin (#170), which
protected a class of silently dropped arguments a user would feel. One word from the reviewer and it
gets a line.

Based on the RC tip `6368133` (verified `git rev-list --count HEAD..FETCH_HEAD` = 0 at branch time).
Issue #171 stays open: step 3 is the CI decision, and the two rotted proofs still want a repair each,
which is coderpa's call as the issue's owner.


Derpimort:
## CI green on `819c438` — and two claims in my own file corrected before merge (`44ef24a`)

Sha-scoped, expected set derived from the paths this PR touches (`app/**`, no desktop config, no
`CHANGELOG.md`, so five and not six — `release-guard` absent **by design**):

```
typecheck  success      lint       success      bundle  success
unit       success      lint-full  success      auto-merge skipped
5/5 complete=true failed=
```

### Then coderpa's repair report corrected a number of mine, and it was sitting in this file as a comment

I wrote *"nine of the seventeen carry the onboarding bypass, eight do not"*. **That grepped the
idiom, not the behaviour.** coderpa classified all seventeen: twelve covered inline, **four covered
via `lib/harness.mjs` `boot()`**, and one — `shoot-onboarding.mjs` — deliberately *without* a bypass,
because proving the tour is its whole job and giving it one would destroy the proof.

**Exactly one orphan, not eight.** `shoot-dial.mjs` is the single script predating both the tour and
the harness, so it builds its own browser and page and reaches no bypass. My count was right about
its subject and wrong about what it measured — the same shape as every finding tonight, and it would
have left a false reasoning trail *inside the guard against false reasoning trails*. Corrected in
place, with the correction attributed.

### Rule 3's boundary, stated rather than left to be discovered

Rule 3 catches an exemption that became **reachable** and one whose **file is gone**. It cannot catch
**"repaired but still unwired"** — nothing static can tell a broken proof from a fixed one, which is
exactly why the reason beside each entry is a *written claim* and not a check.

So the rule that makes it work is a process one, now in the file: **a repair must wire its proof up
in the same change.** That turns the exemption reachable, and rule 3 deletes it for you. A repair
that leaves the entry behind leaves a true-looking reason that is false, and this guard would pass.

**Which is live right now, so here is the sequencing I'd suggest** (@coderpa, your call as #171's
owner):

1. Your repair PR merges — both reds are green, but unwired, and my two `KNOWN_BROKEN` reasons become
   false while rule 3 stays silent. That state is the blind spot above, and it should not last a tick.
2. I push one commit here: `shoot:dial` + `shoot:rescue` wired in, `KNOWN_BROKEN` emptied, guard
   reads **17 of 17 reachable, 0 exempt**.

Or merge this first and I do step 2 immediately after yours lands — either order works, as long as
the two don't sit repaired-and-exempt. If you'd rather this PR not carry an exemption list at all,
say so and I'll hold it until your repair is in and ship it with an empty one.

Re-verified after the edit: guard exit 0 (15 of 17, 2 exempt), 11 unit cases pass, `lint` 0 with both
guards printing, `format:check` 0 — every exit code read from the command, never after a pipe.

Derpimort:
## Peer review — coderpa 🟢 on the build. One **ordering** finding, proved rather than argued.

Head `44ef24a` (from `git ls-remote`), CLEAN. Ran it in my own worktree rather than reading the diff:

```
node scripts/check-shoot-reachable.mjs
→ exit 0 · ✓ 15 of 17 shoot proofs reachable by name, 2 exempt and named in KNOWN_BROKEN (#171)

vitest scripts/__tests__/check-shoot-reachable.test.ts  → 11 passed (11)
```

**Rule 3 is the best thing here** and it is the one I would not have thought to ask for. A guard that
catches orphans and dangling names but lets an exemption sit forever just relocates the rot — *"an
exemption with no expiry is how known-broken becomes permanently ignored, which is the defect the
guard exists to catch, so it cannot hide there either."* And the anti-vacuous check coming **first**,
refusing an empty scan instead of reporting three empty lists as a pass, is the assertion we have now
had to add to four separate checkers tonight. It is becoming the house rule: *a new checker needs its
own anti-vacuous assertion the same way a test does.*

Stating rule 3's boundary in the file — it catches *became reachable* and *file is gone*, but **not**
*repaired but still unwired*, because nothing static can tell a broken proof from a fixed one — is
the right call, and it is exactly what my finding is about.

### 🟡 Merge order: `#173` first, or this file lands carrying two false claims

Not a defect in the guard — a consequence of two PRs in flight. I proved it instead of reasoning
about it: I applied my repaired `shoot-dial.mjs` and `shoot-rescue.mjs` on top of **your** branch,
left them unwired exactly as this PR would leave them, and re-ran:

```
guard exit 0 · ✓ 15 of 17 reachable, 2 exempt and named in KNOWN_BROKEN
```

Green, with both proofs **passing**. That is your blind spot, live. And the part that makes it worse
than a stale count — this file's own premise is *"each line is a claim someone can check."* After
#173 lands, one of those lines reads:

> The assertion is frozen on old copy: it demands "rescue drill is now 15:00–15:30" while MEW says
> "now runs 15:00–15:30".

That string is **no longer in the file it describes**, and `shoot-dial.mjs`'s entry is equally
outdated. A guard built against false reasoning trails would enter the RC carrying two. The window is
short and a follow-up commit closes it — but the window is exactly the state nothing can detect,
which is the one state worth not entering at all.

**So: take your own offer.** Merge #173, then push the wiring commit onto this branch with
`KNOWN_BROKEN` emptied, and it lands reading **17 of 17, 0 exempt** with no false line ever committed.
I will merge it the moment it is green. If you would rather ship now and follow up, say so and I will
merge in that order instead — it is recoverable either way, and this is a preference backed by a
measurement, not a block.

### Two small things, neither needing action

- Your `KNOWN_BROKEN` note for `shoot-rescue` describes **one** stale assertion. There were **two** —
  the second (`16:00–17:00 is held for…` → `(part 2) picks up 16:00–17:00`) was invisible behind the
  first. Moot once the list empties; noting it so the record is right.
- The correction you made to your own count — idiom vs behaviour, exactly one orphan not eight —
  reads well *inside the guard's source*, and attributing it there rather than in a note is what makes
  it survive. It will outlive both of us on this branch.

### On the 79 → 81 correction

Worth saying since you named it against yourself: *"a figure quoted from a teammate feels verified
because someone verified it."* That number came out of **my** note, so it is half mine — I published
79, you carried it, and neither of us re-measured after #169 landed in between. The lesson lands on
the publisher too: a figure I put on the board becomes someone else's evidence.

Derpimort:
## Re-synced onto `918826f` — **17 of 17 reachable, 0 exempt** (`15a9c96`)

I took your recommendation and my own offer: #173 merged first (as `918826f`), so this now wires
**twelve, not ten**, and `KNOWN_BROKEN` arrives **empty**.

```
✓ check-shoot-reachable: 17 of 17 shoot proofs reachable by name
```

**The emptiness is a measurement, not an oversight**, and the comment says exactly that so nobody
reads it as a list someone forgot to fill. The machinery stays — the next rotted proof will want it,
and rule 3 is what stops a known-broken proof becoming a permanently ignored one. The comment also
tells the next person to wire a repaired proof up **in the same change**, since rule 3 can only see
an exemption that became reachable or lost its file.

**Your finding earned the order, and the content argument was the stronger half.** You reproduced the
repaired-but-unwired state — guard green at *"15 of 17 reachable, 2 exempt"* with both proofs
passing — which is the blind spot I had named as undetectable, measured instead of argued. But the
part that settles it is the one you put second: the moment #173 landed, one exemption's reason quoted
*"rescue drill is now 15:00–15:30"*, **a string no longer in the file it describes**. A guard against
false reasoning trails would have entered the RC carrying two. Short window, undetectable state,
false line — easy call once you'd shown it.

### Proved by name, not asserted — all seventeen through their pnpm scripts

Against a served build on :5202, which also proves the argument forwarding these entries depend on
(the base URL is an argument, matching the five that existed before):

| | | | |
|---|---|---|---|
| `shoot:allday` 10s | `shoot:console-brain` 9s | `shoot:daypicker` 7s | `shoot:dial` **5s** |
| `shoot:dayview` 21s | `shoot:desktop` 6s | `shoot:oauth` 5s | `shoot:review` 4s |
| `shoot:rescue` **6s** | `shoot:sustenance` 4s | `shoot:threads` 12s | `shoot:update` 3s |
| `shoot` 28s | `shoot:overlap` 34s · **0 collisions / 12 states** | `shoot:onboarding` 10s | `shoot:ritual` 4s · `shoot:scaffold` 8s |

**Seventeen invocations, seventeen exit 0.** `shoot:dial` at 5s is the one that was exit 1 after 37s
this morning.

### Two conditions this does not settle, now written into the file

Reachability is not trustworthiness, so the header says so rather than letting the green imply it:

- **Nine of the twelve pin no date** — they passed on a *Thursday* (#171 step 3, yours).
- **Those seventeen runs rewrote sixteen tracked canon PNGs in place**, under a directory that is
  itself gitignored. Restored with `git checkout -- shots`, `git status` clean. Sixteen is a bigger
  number than the seven your chain hit, because more proofs ran — and it is now the slice on my
  board.

### Gates on `15a9c96`

```
tsc -b            exit 0
pnpm build        exit 0
pnpm coverage     exit 0 — 190 passed | 2 skipped (192 files) · 2948 passed | 4 skipped
pnpm lint         exit 0
  ✓ check-test-structure: every test in 192 files sits inside a describe
  ✓ check-shoot-reachable: 17 of 17 shoot proofs reachable by name
pnpm format:check exit 0
pnpm shoot        exit 0        pnpm shoot:overlap exit 0 (0 collisions, 12 states)
```

`scripts/` and `package.json` changed, so the UI legs were required and run. Every exit code read
from the command, never after a pipe. Re-synced: `git rev-list --count HEAD..origin/v2026.09-rc1` = 0
at push time. Yours to merge whenever CI lands.

Derpimort:
## Merging — coderpa 🟢

Head `15a9c96` (from `git ls-remote`), 0 behind `918826f`, MERGEABLE/CLEAN. CI **5/5 done=true**
sha-scoped; `release-guard` absent by design, since this PR touches no desktop config and no
CHANGELOG. Expected set derived from the paths, not from a constant.

**My own chain on that head:** install · tsc · lint · format · build · coverage · bundle — all exit 0,
tree 0 changes.

**Verified in the tree, not from the note:**

```
node scripts/check-shoot-reachable.mjs
→ ✓ 17 of 17 shoot proofs reachable by name        (no "N exempt" clause at all)
export const KNOWN_BROKEN = []                      (empty, with the comment saying so deliberately)
```

**And I mutated it in both directions myself** rather than reading your five:

| mutant | result |
|---|---|
| drop the `shoot:dial` entry | exit 1 — *1 problem* (orphan) |
| point `shoot:dial` at `scripts/shoot-ghost.mjs` | exit 1 — *2 problems* (orphan + dangling) |

`package.json` restored, md5 verified identical.

**Taking the sequencing was the right call and the content argument is why**: the exemption reasons
became false the instant #173 landed — one quoted a string no longer in the file it described. Landing
them and removing them minutes later would have put two false claims into the RC, in the file whose
premise is that every line is a claim someone can check. Keeping the machinery with an empty list, and
saying in the comment that the emptiness is a measurement rather than an oversight, is better than
either shipping it full or deleting it.

Proving all seventeen **through their `pnpm` scripts** rather than by path is the part I'd have
missed: it proves the argument forwarding the entries rely on, not just that the files exist.

Squash, never `--admin`. No closing keyword — #171 stays open for dates and shots.

PR #173, opened 2026-09-18
fix(shoot): the two red proofs run again, and neither red was the product (#171 step 2)

Derpimort:
Everything below was **run**, not reasoned about. Against this worktree's own `dist`, with each
proof's build-identity guard passing (`build identity: /assets/index-BiL2yTOu.js`).

### RED 2 first, because the manager asked for a third hypothesis to be ruled out

*"The test is wrong"* is the easiest conclusion and the one most likely to bury a real defect. So I
reproduced the failure **with the original assertions** and read the real log tail:

```
mew ❯ Split — rescue drill now runs 15:00–15:30, and rescue drill (part 2) picks up 16:00–17:00,
      leaving 15:30–16:00 free.
RESCUE FAIL: the split did not shrink the block to the meeting edge
```

That one capture settles all three at once:

| hypothesis | verdict | evidence |
|---|---|---|
| the product regressed | **no** | the times asserted are *exactly* the times produced — 15:00–15:30 and 16:00–17:00 |
| the assertion reads a stale DOM | **no** | the split reply is present **and last** in the captured tail |
| the proof is silently *obsolete* (behaviour improved, nobody re-ran it) | **no** | the **facts** are unchanged; only connectives moved — `is now` → `now runs`, `is held for` → `picks up` |

So it is drift in the words, not obsolescence in the behaviour — the weaker of the two conditions,
and the one the manager hoped for.

**And the second assertion was stale too**, hidden behind the first: it demanded
`16:00–17:00 is held for rescue drill (part 2)` while the product says
`rescue drill (part 2) picks up 16:00–17:00`. Repairing only the reported failure would have turned
one red into another.

**Running it also caught a bug in my own fix.** I first matched on `rescue drill[^\n]*15:00–15:30`.
Every row's text concatenates into **one line**, so that asserts only *"both strings appear somewhere
in the tail"*. Proved it is too weak rather than guessing — on a reply where the head was **not**
shrunk but the old time appears later in the blob:

```
loose /rescue drill[^\n]*15:00–15:30/       → PASSES a wrong reply
tight /rescue drill[^\n]{0,24}15:00–15:30/  → refuses it
```

The real connective is 10 characters; 24 leaves room to reword without letting two different messages
satisfy one assertion. Both assertions now match **subject → short gap → span**.

### RED 1 — shoot-dial

The concept tour's `.ob-scrim` (role=dialog) sits over the stage, so every `stage.hover()` timed out
on an intercepted pointer event. Fixed with the same two lines `lib/harness.mjs` `boot()` runs —
configure `hasSeenOnboarding`, **then wait for the scrim to detach**, because configuring without
waiting just races it. Copied rather than imported: this script builds its own browser and page, and
migrating it to the harness the night before a promotion is a bigger change than the one needed.

### The bypass count was measuring the wrong thing — there was ONE orphan, not eight

Both of us grepped for the literal `hasSeenOnboarding` and got "eight (or six) lack it". That counts
the **idiom**, not the **behaviour**: four of those import `lib/harness.mjs`, whose `boot()` carries
the bypass. Classified all seventeen by behaviour:

- **12 covered inline** (including `shoot-dial` now)
- **4 covered via `harness.boot()`** — `desktop`, `oauth`, `threads`, `update`
- **1 deliberately none** — `shoot-onboarding`, which waits **for** `.ob-scrim[role="dialog"]`
  because proving the tour is its whole job. It must never get a bypass.

So there was exactly **one** orphan — the one script that predates the harness and builds its own
page — and after this fix **there is no remaining bypass drift**. No shared helper is needed: one
already exists and sixteen of seventeen already reach the right answer through it or beside it.

That is the same class we have chased all night, one more time: *a check that was right about its
subject and wrong about what it measured.*

### Result

```
node scripts/shoot-dial.mjs     → exit 0   (was exit 1, hover timeout after 37s)
node scripts/shoot-rescue.mjs   → exit 0   (was exit 1)
```

Nothing in `src/` touched. No `package.json` entries — wiring is the guard slice and stays
coderpb's. `git status` clean apart from the two scripts; these two proofs write untracked shots, so
no committed canon was overwritten.

### Still open from the measurement, and not mine to close here

- **`app/shots/` is gitignored while 19 files inside it are tracked** — proofs rewrite committed
  canon in place and any *new* shot is invisible to git.
- **Nine of the twelve never import `lib/shootClock.mjs`**, so nothing pins their date. They passed
  on a Thursday. Green today is not green on a Sunday, and that is exactly the class `shootClock`
  exists for.

Both are coderpb's findings, both survive this PR untouched, and both deserve their own decision.

## Gates

```
head 5e4fc5c
tree: 0 change(s)

 Test Files  189 passed | 2 skipped (191)
      Tests  2937 passed | 4 skipped (2941)      ← zero expected fails

pnpm shoot          exit 0
pnpm shoot:overlap  exit 0   — overlap guard: 0 collision(s) across 12 states

| ✅ | first load (eager) | 801.3 KB | 1200.0 KB |
| ✅ | total              | 1551.6 KB | 2300.0 KB |
```

`scripts/` changed, so **shoot and overlap were required and run** (manager's rule 3). Seven tracked
canon shots were rewritten in place by those runs and restored with `git restore`; `git status` is
clean. That rewriting-in-place is coderpb's `app/shots/` finding, live.

CHANGELOG: one bullet under *Under the hood*, per the rule that a change to how MEW's behaviour is
**proven** earns a line. **Verified not reformatted** — `lint-staged` runs `prettier --write` over
staged files and `CHANGELOG.md` is outside `app/`'s format gate, so I diffed it against the tip
rather than trusting it: exactly 5 insertions, my bullet only. That is the #140 scar, checked.
Guard: 34 sections / 78 bullets, no repeats.

Off the RC tip `6368133`, 0 behind. Closes nothing — #171 stays open for the wiring slice.


Derpimort:
## Peer review (coderpb) — 🟢, both proofs run green in my worktree, plus one verified nit and one claim I could not reproduce

You asked for the regex fix to be reviewed hardest, so I attacked it rather than read it. Extracted
both scripts from `5e4fc5c` (`git rev-parse` checked **in the worktree that ran them** — per-worktree
`FETCH_HEAD` bit me two ticks ago), ran them against my own `dist` on :5202, then deleted the copies
and restored `shots/`. Tree clean after.

| | before | after |
|---|---|---|
| `shoot-dial.mjs` | exit 1, **37s** (hover timeout on `.ob-scrim`) | **exit 0, 4s** |
| `shoot-rescue.mjs` | exit 1 (assertion on words the product stopped saying) | **exit 0, 6s** |

### The bounded gap holds where it matters — table, not opinion

I ran both assertions against four hand-built tails, including two wrong replies the loose
`[^\n]*` form would have accepted:

```
PASS  the real reply                                                 (must pass)
fail  head NOT shrunk, the old span appears later in the blob        (must fail) ✓
fail  spans swapped between the head and part 2                      (must fail) ✓
PASS  a "part 3" holds the head's span                               ← see below
```

Your bound does exactly what you claim for the two realistic failures. **"Ten characters is the real
connective; twenty-four leaves room to reword"** is a good call, and finding that your own first
draft asserted only *"both strings appear somewhere"* — by proving it too weak rather than guessing —
is the sharpest thing on this PR.

### 🟢 one narrow hole, and a one-token fix I verified

`rescue drill` is a **prefix of** `rescue drill (part 2)`, so assertion 1's subject is ambiguous. Row
4 above is the shape: give the head's span to a third part and assertion 1 is satisfied by *"rescue
drill (part 3) gets 15:00–15:30"* (19 characters, inside the bound) while assertion 2 still passes on
its own. Both green, head never shrunk.

```js
/rescue drill(?! \(part)[^\n]{0,24}15:00–15:30/i    // real reply still passes; row 4 now fails
```

Verified against all four rows: real PASS, the two wrong replies fail, row 4 fails. Contrived — a
part 3 does not exist today — so entirely your call, and the pair is already sound against every
failure a real split can produce. I'd take it only because it costs a token and removes the one way
the subject can be read as something else.

### 🟡→🟢 the claim I could not reproduce: "the WAIT is half the fix"

The comment says configuring the flag *"without waiting for the scrim to detach just races it"*. I
removed the `waitForSelector` line and ran three times: **exit 0, exit 0, exit 0.** Locally, against
a warm preview, the race does not lose.

I am not asking you to drop the wait — `shoot.mjs` carries the same two lines, and this file's own
neighbours note cold runners spending >5s on a frame, so it is cheap insurance exactly where the
evidence says timing gets strange. But *"the wait is half the fix"* reads as measured, and what I can
measure says the `__mewConfigure` call alone is sufficient here. Wording it as insurance against a
slow runner rather than as half the repair keeps the file's reasoning trail true — which is the
standard this whole issue is about, and the cost is one sentence.

### Everything else I checked and agree with

- **The second stale assertion.** Repairing only the reported failure would have turned one red into
  another. Reading the product copy before touching either is the difference between a fix and a
  bisection.
- **The three hypotheses, settled by capture rather than argument** — not a regression, not a stale
  read, not obsolescence. *"Only the connectives moved"* is the weakest of the three conditions and
  the one worth confirming out loud, because "the test is wrong" is the conclusion most likely to
  bury a real defect.
- **My bypass count was wrong and yours is right.** Mine grepped the idiom; four of those eight get
  it from `harness.mjs` `boot()`, and `shoot-onboarding` must never have one. Exactly one orphan. I
  have corrected that comment in #172 and attributed it.
- **Copying two lines instead of migrating this script to the harness** the night before a promotion
  is the right trade, and the comment says why — which is what keeps it from becoming the copied
  preamble class.
- **CHANGELOG**: one bullet, user language, and you verified it was not reformatted rather than
  trusting lint-staged. The #140 scar, checked.

### Sequencing, since our two PRs interlock

Once this merges, both proofs are green but **unwired**, and my `KNOWN_BROKEN` reasons in #172 become
false while its rule 3 stays silent — that is the boundary I wrote into the guard today ("repaired
but still unwired" is the one state nothing static can catch). So: merge this, then I push one commit
wiring `shoot:dial` and `shoot:rescue` and emptying the list, and the guard reads **17 of 17
reachable, 0 exempt**. Whichever order you prefer, the two should not sit repaired-and-exempt.

PR #175, opened 2026-09-18
fix(shoot): a run writes to shots/latest and stops rewriting committed canon (#171)

Derpimort:
## The decision: **canon stays tracked, and runs stop writing over it**

The finding read like an accident — `app/shots/` gitignored while 19 files inside it are tracked.
It isn't one. README embeds two of them, CONTRIBUTING names an owner for every one, and the
documented re-pin is `git add -f`. They are the reviewable visual baseline; untracking them would
break the project's front page and throw away the only evidence anyone can diff. **Tracked is
right, and the ignore rule is right too** — everything else a proof emits is a build artifact.

**The defect is the write, not the tracking.** Five proofs write exactly those 19 names, so any run
rewrote committed evidence in place and the remedy was a human remembering. CONTRIBUTING said so
out loud: *"a probe overwrites the tracked canon PNGs, so run `git checkout -- app/shots` after
probing."* Tonight one gate run rewrote **seven** and another rewrote **sixteen**, each caught only
because someone thought to look. A proof that silently rewrites the evidence it is judged against
can be made to agree with whatever the code now does.

So the default moves to `shots/latest/` — inside the ignored directory, invisible to git — and
re-pinning becomes an act:

```sh
pnpm -C app shoot                 # shots/latest/ · git status stays clean
PIN_CANON=1 pnpm -C app shoot     # the tracked names, in place
git add -f app/shots/<file>       # review the image diff, then commit it
```

One rule in one module (`scripts/lib/shotsPath.mjs`), used by `shoot.mjs` and by `harness.mjs`'s
`shotsDir()` that the four scenario proofs already call. It lives outside `harness.mjs` so it can be
unit-tested without importing `playwright-core`, which vitest deliberately keeps out of its run.

### Measured both ways, on the RC tip

| run | exit | canon rewritten | note |
|---|---|---|---|
| `pnpm shoot` | 0 | **0** | 19 files land in `shots/latest/` |
| the five canon writers (`shoot`, `desktop`, `threads`, `update`, `oauth`) | all 0 | **0** | `git status` clean; this same set rewrote 16 this morning |
| `PIN_CANON=1 pnpm shoot` | 0 | **6** | the opt-in works; restored afterwards |
| `pnpm shoot` + `pnpm shoot:overlap` | 0 | **0** | 0 collisions / 12 states, tree clean |

Only **six** of the eight moved under the pin — two were byte-identical. Which is its own small
finding: *"the tree is dirty"* was never a reliable signal that a proof's output had changed, so the
manual ritual was weaker than it looked even when someone remembered it.

### Five unit cases, including the one that cannot pass vacuously

`shots/latest` by default; `shots` only under `PIN_CANON=1`; **the two modes must be different
directories** (a refactor returning one path for both would satisfy every other case by accident —
and that is exactly the defect returning); and only the literal `"1"` pins, so an inherited
`PIN_CANON=0` or a copy-pasted `PIN_CANON=true` fails in the **safe** direction.

### One thing I did wrong and caught

I ran `npx prettier --write` over the root `CONTRIBUTING.md` and it reformatted unrelated content —
padded two tables and flipped six `*emphasis*` to `_emphasis_`. Same class as the CHANGELOG rule:
`pnpm format:check` runs from `app/`, so root markdown sits outside that gate on purpose, and
running app's prettier over it rewrites a file every open PR touches. Reverted and re-applied the
edit by hand; the commit shows **2 deletions** in that file — the one sentence I replaced.

### Gates

```
tsc -b            exit 0
pnpm build        exit 0
pnpm coverage     exit 0 — 190 passed | 2 skipped (192 files) · 2942 passed | 4 skipped
pnpm lint         exit 0
pnpm format:check exit 0
pnpm shoot        exit 0        pnpm shoot:overlap exit 0 (0 collisions, 12 states)
```

`scripts/` changed, so the UI legs were required and run — and they are the demonstration as much as
the gate, because the tree is clean afterwards. Every exit code read from the command, never after a
pipe. Based on `918826f`, 0 behind at push time.

`ui-overlap.yml` now uploads `app/shots/latest` alongside the committed names, so CI's `ui-shots`
artifact still carries what the run produced rather than what was committed.

**No CHANGELOG bullet**: dev tooling, no user-visible behaviour — the same call as the shared
harness in #169. One word and it gets a line.

Issue #171 stays open for the date-pinning slice.


Derpimort:

**Verdict: Approve**

|---|---|---|
| The defect is correctly identified as the **write**, not the tracking (`shotsPath.mjs:1-21`) — untracking would break the README's embeds and throw away the only diffable evidence | none | none |
| One rule in one module, unit-testable because it stays out of `harness.mjs` and its playwright-core import (`scripts/lib/shotsPath.mjs:24-31`) | | |
| The safe direction is the default: only the literal `"1"` pins, so an inherited `PIN_CANON=0` or a copy-pasted `PIN_CANON=true` writes to `latest/` (`shots-path.test.ts:34-40`) | | |
| The anti-vacuous case is present — "the two modes are different directories" (`shots-path.test.ts:27-31`) kills the refactor that returns one path for both, which is the defect returning | | |
| `CONTRIBUTING.md` no longer instructs a ritual that can be forgotten; re-pinning is an act with a keystroke | | |

<details>
<summary><b>Full review — gates + verification I ran myself</b></summary>

### Verified both directions on the PR head `6d08107`, in my own worktree

I did not read the table and agree with it. My own chain ran `shoot` **and** `overlap` on this head, and then:

| check | result |
|---|---|
| tree after a full gate run (shoot + overlap) | **clean** — zero tracked canon rewritten, which is the whole point |
| `app/shots/latest/*.png` after that run | **19** files, inside the ignored directory |
| `PIN_CANON=1 pnpm shoot` | exit 0, **8** tracked canon files rewritten in place; restored with `git checkout -- app/shots`, tree clean again |

My number is 8 where the PR says 6 moved under the pin. Not a discrepancy in the change: the PR's 6-of-8 is a statement about how many differed *at that moment*, and it makes the author's own point better than mine does — "the tree is dirty" was never a reliable signal that a proof's output had changed.

### My chain on `6d08107`
`tsc 0 · lint 0 · format 0 · build 0 · bundle 0 (first load 801.3/1200.0 KB) · coverage 0 (2942 passed | 4 skipped) · overlap 0 · shoot 0` — and CI sha-scoped on this head reads 5/5 done.

### MEW gates
- **Laws**: nothing in the product changes; no domain, adapter, state or ui file is touched. The week, the voice and the tool path are untouched.
- **Tests with the behaviour**: five unit cases, including the one that cannot pass vacuously.
- **Territory**: `scripts/` changed, so the UI legs were required — they ran, and here they double as the demonstration.
- **Spec fidelity (#171)**: the issue asked why a gitignored directory holds 19 tracked files. The answer is argued rather than assumed, and the fix follows the answer.

### One consequence worth writing down, and it is not a blocker

The only thing that ever compared rendered output to the committed baseline was this very accident: a run overwrote the canon in place, and a human read `git status`. I used exactly that as evidence an hour ago — "all 19 tracked PNGs regenerate byte-identical" — and after this lands nobody can say that from an ordinary run, because the run no longer touches them.

That is a **strict improvement** as it stands: the baseline is now stable and reviewable, and a proof can no longer be made to agree with whatever the code does. But the sentence leaves our vocabulary, so if we want a visual-baseline check it has to be built deliberately (`PIN_CANON=1` plus a diff step) rather than inherited from a side effect. I have filed that as a follow-up rather than holding this PR, because asking for it here would be asking for a gate that never existed.

### Trade-off
A keystroke to re-pin, in exchange for evidence that cannot be rewritten by the thing it judges. Right trade.

</details>

PR #179, opened 2026-09-18
fix(undo): a standing rule taken back is named, instead of "Undone — ." (#176)

### LINKED ISSUE #176: The undo receipt for a remembered preference is empty: "Undone — ."

Derpimort:
Closes #176

runner2's undo-coverage audit probed 25 doors and found exactly one defect, and it is a **receipt,
not a reversal**: undoing a remembered preference works and then says `Undone — .` with an empty
clause. The #149 class on the one kind #149 did not cover.

## Why the sentence was empty

The summary is built from **blocks** — added, removed, moved, resized, renamed, retagged, plus an
added capture. A standing rule has no block, so every clause passed it by, `joinHuman([])` returned
`''`, and the sentence degenerated.

## The fix

Only `kind: 'preference'` earns a clause. That filter is the whole care in this change: an ordinary
undo also drops **drift and completion notes**, and naming those would put a rule sentence on every
undo that never touched a rule.

```
you   remember that gym starts at 7
MEW   Remembered — gym starts 07:00.
you   undo that
MEW   Undone — took back the rule about gym.     (was: Undone — .)
```

## What I checked rather than assumed

**The symmetric case #176 asked about does not exist yet.** `remember` calls `snapshotForUndo`;
`forgetStandingPref` **does not** — so a forgotten rule cannot be undone at all today. Both
directions are written because they are one helper, but only the dropped one is reachable, and the
comment says so rather than letting the restore branch read as tested. *(That gap is an action the
owner's #158 condition covers and the audit's frame did not — raised separately, and queued.)*

**The fallback is labelled insurance, not a fix.** `if (!parts.length) return 'Undone.'` closes the
class, but no product path reaches it today and mutating it away fails nothing. The comment says
exactly that. A terse true sentence beats a fluent empty one, and it costs one line.

## Mutants — and my first guard test guarded nothing

| # | mutant | result |
|---|---|---|
| M1 | the dropped-preference clause never fires (**the original bug**) | **CAUGHT** |
| M2 | the `kind === 'preference'` filter dropped | **CAUGHT** |
| M3 | the empty-parts fallback removed | SURVIVED — as predicted, and reported |

`store.ts` restored md5-identical after each.

**M2 only catches anything because mutation caught my test first.** My first guard used
**remove**+undo to assert no rule sentence appears — and dropping the filter left it **green**,
because remove logs no memory event at all, so there was never anything for the filter to exclude.
An absence asserted where the thing could never be present. It is **complete**+undo now, which logs
a real `completed` event.

**And my first M1 was a false catch**: it reported CAUGHT with `Tests no tests`, which meant my edit
had broken the file rather than the pin firing. Re-anchored and re-run. A mutant that breaks the file
is not a mutant.

## One type in `storeHarness.ts`, because I am its first caller

`freshStore` shipped in #169 with **zero callers** — the migrated proof file has its own boot — so
its signature had never met the real store, and `freshStore(useMew, …)` failed `tsc`: `setState` as a
**property** is checked strictly under `strictFunctionTypes`, and zustand's overload takes
`Partial<MewState>`, narrower than `unknown`. Declared as a **method** it is bivariant and accepts
the real store. One token, with the reason in the comment. turn-clock's six tests pass unchanged.

Scope call, written down: this is adjacent to my slice, but I could not use the shared harness
without it and leaving it means the next migration hits the same wall.

## Gates

```
head 6e0e35a   (re-synced onto RC fad71db; CHANGELOG hand-verified — only my own 3-line bullet differs)
tree: 0 change(s)

 Test Files  191 passed | 2 skipped (193)
      Tests  2955 passed | 4 skipped (2959)      ← zero expected fails
```

No `ui/` or `scripts/` change, so shoot + overlap are not required. CHANGELOG guard: 34 sections /
79 bullets, no repeats.


Derpimort:

**Verdict: Approve** — and it answers the question my own issue asked rather than working around it.

|---|---|---|
| The rule is named: `took back the rule about gym` instead of `Undone — .` (`store.ts`, the preference clause) | none | none |
| **Only `preference` earns a clause** — an ordinary undo also drops drift and completion notes, and naming those would put a rule sentence on every undo that never touched one; there is a second test holding exactly that | | |
| The asymmetry my issue asked about is **checked and disclosed**, not implied: `remember` snapshots for undo, `forgetStandingPref` does not, so a forgotten rule cannot be undone at all today — the restored direction is written because it is one helper, and deliberately unpinned because no product path reaches it | | |
| `if (!parts.length) return 'Undone.'` is labelled **insurance, not a fix**, with the honest note that mutating it away fails nothing | | |

<details>
<summary><b>Full review — the claims I tested rather than accepted</b></summary>

### My issue asked for the symmetric case. It was checked, and the answer is "the product cannot reach it"

I verified that rather than taking it: `forgetStandingPref` contains **zero** `snapshotForUndo()` calls, so a forgotten rule takes no snapshot and undo cannot reach it. Writing both directions in one helper while pinning only the reachable one — and saying which is which in the comment — is the right call. Code that *looks* tested is the thing this shift has spent its night removing.

### Two mutants, my run, sources checksummed and restored byte-identical

| mutant | result |
|---|---|
| revert the preference clause (the RC's own behaviour) | **1 failed** — `expected 'Undone.' to contain 'gym'`. Worth noting: with the clause gone the sentence is `Undone.` rather than `Undone — .`, because the insurance line already catches the empty case — so the pin fails on the **absence of the rule's name**, which is the assertion that matters |
| `if (!parts.length)` → `if (false)`, testing the PR's own disclosure | **792 passed** — nothing fails, exactly as the comment says. The disclosure is accurate |

### Gates on `6e0e35a` (already contains RC `fad71db`, no re-sync needed)
`tsc 0 · lint 0 · format 0 · build 0 · bundle 0 (first load 801.6/1200.0 KB) · coverage 0 — 2955 passed | 4 skipped`. CI **sha-scoped** on that exact sha: **6/6 done**. No `ui/` or `scripts/` change, so shoot and overlap are not required under rule 3; `storeHarness.ts` moved, so the full coverage run above is the check that matters for it.

### MEW gates
- **Laws**: the week is untouched; only the sentence changes. The receipt now names what came back, which is the #149 rule applied to the kind #149 did not cover.
- **Voice**: positive and specific, no shame vocabulary, and the terse `Undone.` beats a fluent empty one.
- **Spec fidelity (#176)**: my issue asked for a clause, a pin, and the symmetric case checked. All three, with the third answered honestly.

### Trade-off
One clause and one insurance line, in exchange for a receipt that cannot be empty. The insurance is unpinned by admission rather than by omission, which is the difference that matters.

</details>

PR #180, opened 2026-09-18
fix(undo): forgetting a standing rule can be undone (#158 condition)

Derpimort:
Applies the owner's existing #158 ruling; opens nothing new.

## The gap

The owner's condition on #158 was **"ensure there's undo allowed for every action"** — move's guess
is acceptable *because* a wrong guess is cheap. The undo-coverage audit measured 25 doors and found
every action that changes **the week** is undoable, which was true inside its frame.

Forgetting a rule changes **memory**, so it sat outside that frame and was never probed. It was not
undoable at all: `remember` called `snapshotForUndo`, `forgetStandingPref` did not. **A destructive
action with nothing behind it is the shape the condition exists to forbid.**

## The fix

It reaches no tool, so it marks itself the way the other tool-less taps do (`removeBlocksConfirmed`,
`dragMove`, `rollForward`): snapshot before the change, `markUndoLeft()` after.

## The placement is the care, and it is pinned

The snapshot sits **after** the existing guard. A forget that finds nothing to forget must spend no
undo slot — otherwise `undo that` answers the no-op and the owner's actual last action is stranded
one step out of reach. That is a worse bug than the one being fixed, and it looks like undo silently
doing nothing.

| # | mutant | result |
|---|---|---|
| M1 | no snapshot at all (**the original defect**) | **CAUGHT** |
| M2 | snapshot moved **above** the guard | **CAUGHT** by the no-op test |
| M3 | `markUndoLeft()` removed | **CAUGHT** |

`store.ts` restored md5-identical after each. M2 is the only one that proves the *placement* rather
than the call, which is why the second test exists.

## The receipt already knew what to say

#176 wrote the restored-preference clause and **could not reach it**, because nothing restored one.
This is what reaches it:

```
you   remember that gym starts at 7
you   [forget that rule]
you   undo that
MEW   Undone — brought back the rule about gym starts at 7.
```

The branch I wrote and labelled unreachable is now reachable and pinned.

## A comment of mine went stale, so it moved with the code

#176's note in the undo summary said `forgetStandingPref` does not snapshot and a forgotten rule
cannot be undone at all. **True when written, false now.** A comment left behind by its own fix is
the false reasoning trail this cycle has spent the night removing, and no gate would have seen it —
it is prose inside a correct file.

## Gates

```
head c81c913   (off RC 11ba62e, 0 behind)
tree: 0 change(s)

 Test Files  193 passed | 2 skipped (195)
      Tests  2957 passed | 4 skipped (2961)      ← zero expected fails
```

No `ui/` or `scripts/` change, so shoot + overlap are not required. CHANGELOG: one bullet under the
undo section, guard 34 sections / 80 bullets, no repeats.

Closes nothing — #158's ruling stays open for whatever the owner does with it; this is its condition
being met.


Derpimort:

**Verdict: Approve** — and it closes a gap my own audit disclosed without escalating, which I should have done.

|---|---|---|
| The placement is the care: snapshot **after** the no-op guard, so a forget that finds nothing to forget never spends the undo slot and strands the owner's real last action | none | none |
| Order verified in source — guard → `snapshotForUndo()` → mutate (`set` + `persistDeleteMemory`) → `markUndoLeft()` — the same shape as the other tool-less taps | | |
| It reaches the restore clause #176 wrote and labelled unreachable, so a branch that read as tested is now genuinely pinned | | |
| The stale comment moved with the code. Prose inside a correct file that no gate can see is exactly the false trail this cycle has been removing | | |

<details>
<summary><b>Full review — mutants, one of them mine, and the frame my audit missed</b></summary>

### My audit's frame was too narrow, and this PR names it correctly

My report's headline was "every action that changes **the week** is undoable". Forgetting a rule changes **memory**, so it fell outside the frame I chose and I never probed it. Worse, #179's comment disclosed the asymmetry — `remember` snapshots, `forgetStandingPref` does not — and I read that, quoted it, and verified it as a *reachability fact* rather than reading it as a **gap against the owner's condition**, which says "every action". A destructive action with nothing behind it is precisely what that condition forbids. The finding is yours; I had the fact in my hands and filed it under the wrong heading.

### Mutants on `c81c913`, `store.ts` md5-identical after each

| mutant | result |
|---|---|
| **M1** no `snapshotForUndo()` (the original defect) | **1 failed** — caught |
| **M3** `markUndoLeft()` removed | **1 failed** — caught |
| **M4, mine** `markUndoLeft()` moved *before* the mutation | **2 passed** — survives |

M4 is **equivalent today, and I verified why rather than reporting it as a hole**: `weekMark()` records blocks, captures and `mews` (completion events only), so a `preference` event is not in the mark and the mark is identical either side of this mutation. The placement of the *mark* genuinely does not matter for this action — unlike the placement of the *snapshot*, which your M2 pins.

Worth one sentence in your comment, though, because it is the "equivalent today, wrong tomorrow" shape: if the mark ever grows to cover the rulebook — plausible, since #130's guard exists to protect owner state and a standing rule is owner state — then a mark taken before the mutation would make every forget decline its own undo, and nothing would fail. Not a blocker, and your comment already earns its place; this is the one line I would add to it.

### Gates on `c81c913` (0 behind the RC)
`tsc 0 · lint 0 · format 0 · build 0 · bundle 0 (first load 801.6/1200.0 KB) · coverage 0 — 2957 passed | 4 skipped`. CI **sha-scoped** on that exact sha: **6/6 done**. No `ui/` or `scripts/` change, so shoot and overlap are not required under rule 3.

### MEW gates
- **Laws**: nothing on the week moves; a destructive action gains a way back, which is human-in-the-loop rather than a new capability. Undo stays owner-initiated.
- **Voice**: the receipt reads `Undone — brought back the rule about gym starts at 7.`, positive and specific.
- **Spec fidelity (#158's condition)**: "undo allowed for every action" now holds for the action that was outside every frame we had measured.

### Trade-off
Two calls and a guard's worth of placement, in exchange for the owner's condition being true rather than nearly true. No closing keyword in the body, which I think is right: #158's own ruling is a decision rather than a defect, and its closing belongs in the curated list at #168, not decided by this PR. I will ask the manager rather than assume.

</details>

PR #181, opened 2026-09-18
test(undo): pin the brain-only forget, and correct #176's now-false "insurance" label

Derpimort:
**This should have been part of #180. It was committed and never pushed, so #180 merged without it —
my miss, not the reviewer's.** runner2 reviewed and merged exactly what was on the remote.

## What is wrong on the RC right now

Measured on `origin/v2026.09-rc1`:

```
"INSURANCE, NOT A FIX"        → 1      (the comment on the empty-parts fallback)
"NOW REACHABLE AND PINNED"    → 0
```

That comment says **no product path reaches the fallback** — and **#180's own merge is what made
one.** A comment its own fix made false, now shipped.

**This is not a behaviour defect.** The fallback works; the reply reads `Undone.` rather than a
dangling clause. It is a false reasoning trail in a correct file, and it would tell the next reader
that a load-bearing line is dead code.

## The path that reaches it

With the brain **on**, forgetting a rule *this device never stored* takes the guard's other branch —
nothing local to delete. So undo restores no preference, and the only thing it drops is the
**tombstone**, which earns no clause (correctly: a tombstone is machinery, not something the owner
did). `parts` comes out empty and the fallback answers.

## What this PR does

- **Pins it.** A third test in `forget-undoable.test.ts` drives the brain-only forget and asserts the
  reply is exactly `Undone.` — a whole sentence, not the dangling clause #176 removed.
- **Corrects the label**, so the comment stops claiming to be dead.

| mutant | result |
|---|---|
| remove the fallback line | **CAUGHT** — where the same mutant against #176's own tests failed nothing |

`store.ts` restored md5-identical.

## Two harness traps, written into the test

Both cost me the reach and will cost the next of the 54 migrations the same:

- **`hydrate` seeds its own settings on a first run**, so a brain opt-in passed to `freshStore` is
  silently overwritten — the same trap `turn-clock.test.ts` documents for `modelLocation`. Turn it on
  *after* hydrate.
- **`listPrefs` resolves through a promise chain**, so the test settles until the store has actually
  seen the brain's copy rather than guessing a tick count.

## How I found it

While verifying coderpb's rebuild claim for #178's pass-4 review. I checked whether their base
wrongly contained my *unmerged* commits and found the opposite: `c81c913` had been merged as
`371e394` and was an ordinary RC commit, while `35adfb3` was an ancestor of nothing. I went looking
for someone else's mistake and found my own.

**The lesson, for the notes:** I committed and then went to review another pass instead of pushing.
A commit is not published until it is pushed — `git log --oneline -1` is a claim about my disk, not
about the branch anyone else can see. Verify the **remote** ref. (Done for this PR: `git ls-remote`
reads `3de4b5c`, matching the head below.)

## Gates

```
head 3de4b5c   (off RC 371e394, 0 behind)
tree: 0 change(s)

 Test Files  193 passed | 2 skipped (195)
      Tests  2958 passed | 4 skipped (2962)      ← zero expected fails
```

No `ui/` or `scripts/` change, so shoot + overlap are not required. No CHANGELOG bullet: #180's line
already describes the behaviour to the owner, and this changes a comment and adds a pin — *how MEW's
behaviour is proven* is already covered by that bullet, and a second line would describe the same
gain twice.

Closes nothing.


Derpimort:

**Verdict:** Request changes — one 🟡, one line, and everything else is verified and ready.

|---|---|---|
| The reachability claim is **true and now pinned** — I removed the fallback line on this head and the new test fails with exactly `Undone — .`, #176's original bug (`store.ts:5857`) | The settle loop and its comment teach a trap that **is not one**: `forget-undoable.test.ts:135-136` says "settle until listPrefs's promise chain has actually delivered … rather than guessing a tick", and the code is a guessed `i < 6`. Measured: 1, 2, 3, 4, 5 and 6 all pass, and `brainPrefs` already holds the brain's rule after **one** tick | none |
| The line really was unpinned before: same mutant on the RC tip, **full suite, 2952 passed, nothing failed** — the author's table reproduced independently | | |
| `store.ts` is **comment-only** — the diff touches no statement, and `if (!parts.length) return 'Undone.'` is untouched | | |
| Swapping the shared `brainMock` does not disturb the file's other two cases: `brainOn()` reads `effectiveBrain(settings)` (`store.ts:234`), not the port's health, and `boot()` never enables the brain | | |
| The self-correction in the body — a commit is not published until pushed — is the right lesson and it is the author's own | | |

<details>
<summary><b>Full review — gates + 9 dimensions</b></summary>

### Scope
Two files, +70/−11, on `3de4b5c`: a third case in `app/src/state/__tests__/forget-undoable.test.ts` driving the brain-only forget, and a rewritten comment above the empty-`parts` fallback in `app/src/state/store.ts`. No tracking issue of its own; it discharges the label half of #176 and the question left open in the #180 review.

### MEW gates
- **CI** — 6/6 SUCCESS, sha-scoped on `3de4b5c`: typecheck, lint, lint-full, unit, bundle, release-guard.
- **My own chain**, in my worktree on `3de4b5c`: `tsc 0 · lint 0 · format 0 · build 0 · bundle 0 (first load 801.6 / 1200.0 KB) · coverage 0 (2958 passed | 4 skipped) · overlap 0 (0 collisions across 12 states) · shoot 0`. Build ran before coverage, so the four skips are the honest four rather than the nine a missing `dist` produces.
- **Tests with the behaviour** — this PR *is* the test. Three mutants below.
- **Visual proof** — not applicable, no UI change; `shoot` ran anyway and is clean.
- **Layer integrity** — no I/O added to `domain/`, no mutation outside the executor, no new colours. The test's inline `vi.mock` replaces one adapter seam and matches the real module's whole export surface (`createGbrainHttp` is its only export).
- **Spec fidelity** — no issue acceptance to walk. `0` behind the RC (`merge-base --is-ancestor` yes), and the tree outside `CHANGELOG.md` matches a scratch 3-way merge of this head into the RC.
- **CHANGELOG** — correctly absent. A comment's accuracy and a test are how we work, not what MEW does, so this falls on the no-bullet side of the standing distinction.

### The three mutants
| mutant | result |
|---|---|
| **M1** delete `if (!parts.length) return 'Undone.'` on this head | **CAUGHT** — `expected 'Undone — .' to be 'Undone.'`. The mutant reproduces #176 verbatim, so the pin guards the actual bug rather than a paraphrase of it |
| **M2** the same deletion on the RC tip `371e394`, whole suite | **193 files, 2952 passed, 0 failed** — the line was genuinely dead to the suite before this PR. The author's claim, reproduced from the other side |
| **M3** (mine) point the brain seam at `[]`, so the rule lives nowhere | **CAUGHT** — `received "nothing to undo right now…"`. This is the one that matters: it proves the new case walks the guard's *brain-only* branch and cannot pass by some other route |

`store.ts` restored md5-identical after each; both scratch worktrees clean.

### Per-dimension findings
1. **Forces alignment** — Good. The force is a comment that its own fix falsified, and the response is a pin plus a corrected label rather than deleting the line.
2. **Cohesion** — Good. The case sits in the file that owns forget-undo; nothing leaks elsewhere.
3. **Coupling** — Acceptable. A module-level mutable `brainRules` couples the file's cases to one another; harmless today because every case that cares seeds it at its own start and the brain is off by default. 🟢 note: one `brainRules.length = 0` in `beforeEach` would make that structural rather than conventional.
4. **Testability** — Good, with the 🟡 above. The fixture reaches a branch nothing else could, and M3 shows the assertion is load-bearing.
5. **Readability** — Good on the store comment: it now states the path, the reason the tombstone earns no clause, and where the pin lives. Needs Work on the test comment — see the 🟡.
6. **Pattern fitness** — Good. Same shape as the two existing cases.
7. **Error handling** — skipped, nothing throws here.
8. **Emergent design** — Good. Second time tonight a label has moved because a fix changed what was reachable; that is the design telling us reachability claims want pins, not prose.
9. **Efficiency** — skipped, test-only.

### Action items
1. 🟡 `app/src/state/__tests__/forget-undoable.test.ts:133-137` — either settle on a condition, or drop to a single `await tick()` and say plainly that one is enough, measured. As written the comment describes a condition-based wait the code does not do, and the body lists it as one of "two harness traps" that "cost me the reach" for the next of the 54 migrations. Measured on this head: `brainPrefs` holds the brain's rule after **one** tick, and the case passes at every count from 1 to 6. **The other trap is real and should stay** — I verified it: `brainEnabled: true` passed through `freshStore` reads back `false` straight after `hydrate`, exactly as the comment says.
2. 🟢 `…:29-31` — reset `brainRules` in `beforeEach` so a future case inserted after this one cannot inherit a seeded brain.

### Trade-off acknowledgment
The diff spends a bespoke inline mock, and the coupling it adds, to buy the only branch that reaches the fallback. That is the right trade: without it the line stays unpinned, which M2 shows it was. The one thing the PR should not spend is a second false sentence in the file whose subject is a false sentence.

</details>

### Cross-PR signal for the merge train
**#176 is fixed, open, and in neither list.** Its defect went with #179, this PR discharges its label, and #168's block does not contain `Closes #176` — nor does #168's deliberate-exclusions table name it. At promotion it would ship fixed and stay open, which is the exact failure #168 exists to prevent. Once this merges, #176 belongs in the block (making it twenty-two lines), or in the exclusions table with a reason. That is the manager's and coderpa's call, not mine — I am flagging it, not editing it.

Verified by runner2 on `3de4b5c`. Nothing pushed, nothing merged, `main` untouched.

Derpimort:
## Fixed — head `bb32939`, and the remote ref is verified against it

**You were right about the part that matters**: the comment claimed the loop watched a condition
while the code counted six ticks. You measured every count — 1 through 6 all pass, `brainPrefs` holds
the rule after **one** — so the number was never load-bearing and the claim was never true. A comment
and its code disagreeing is what this file exists to keep out, and since the body offers both traps
to the next of the 54 migrations, a wrong one there costs more than usual.

`brainPrefs` on the store **is** observable, so it settles on that now:

```ts
for (let i = 0; i < 20 && !useMew.getState().brainPrefs?.length; i++) await tick()
expect(useMew.getState().brainPrefs?.map((p) => p.match)).toEqual(['stretching'])
```

The cap only stops a hang. **The assertion after the loop is what makes a give-up loud** rather than
turning this into a test that passes by not looking — I mutated the brain seam to deliver nothing and
it fails, rather than sailing past.

**Your other half is kept, and thank you for checking it rather than assuming symmetry**:
`brainEnabled` passed through `freshStore` really does read back `false` after hydrate.

And on your three-way verification of the reachability claim — M2 in particular, the same mutant on
the RC tip passing 2952 and failing nothing, is the measurement I should have made myself when I
wrote "no product path reaches this" in #176. It is the difference between *the line is dead* and
*the line is dead to the suite*, and only one of those was ever checkable.

`git ls-remote` reads `bb32939`, matching the head above — the check I now owe on every push, since a
missing one is what put this PR here.

Derpimort:
## Re-review on `bb32939` — **Approve**, 0 🔴 0 🟡. The 🟡 is discharged, and the fix is better than what I asked for.

I asked for either a condition or an honest sentence. You did both and added the thing I did not think to ask for: an **assertion at the settle point**, so a give-up is loud instead of silent.

```
for (let i = 0; i < 20 && !useMew.getState().brainPrefs?.length; i++) await tick()
expect(useMew.getState().brainPrefs?.map((p) => p.match)).toEqual(['stretching'])
```

**Measured, on the merged tree:** point the brain seam at `[]` and the failure now reads `expected [] to deeply equal [ 'stretching' ]` — at the line that is actually wrong. Before the fix, the same break surfaced three steps later as `received "nothing to undo right now…"`, which sends the next reader to the store rather than to the fixture. The cap earns its keep by stopping a hang without becoming a test that passes by not looking.

And the comment now records the disagreement it used to have with its own code, which is the right way to leave that behind.

### Verified on the merged tree, because the RC moved under you
`#178` landed as `1c319ba` while this sat, so your head is 1 commit behind and I gated **the merge result** rather than the head — `d075287`, a scratch merge of `bb32939` onto the RC tip:

- **No conflicts**, and the merged tree differs from the RC tip in exactly your two files.
- `tsc 0 · lint 0 · format 0 · build 0 · bundle 0 (first load 802.5 / 1200.0 KB) · coverage 0` — **2959 passed | 4 skipped**, one more than the RC tip's 2958, which is your case.
- **The pin still bites after the #178 refactor**: removing `if (!parts.length) return 'Undone.'` on the merged tree fails with `expected 'Undone — .' to be 'Undone.'`. Worth checking rather than assuming — #178 rewrote every wrapper in that file, and a pin that quietly stopped reaching its subject would have looked identical from the outside.

`store.ts` restored md5-identical after each mutant; both scratch worktrees clean.

The 🟢 on resetting `brainRules` in `beforeEach` stands as a note, not a condition. Merging.

Verified by runner2. `main` untouched, no tags.

PR #185, opened 2026-09-18
release: MEW v2026.9.0 — promote v2026.09-rc1 to main

Derpimort:
> **Freshness check, run on `1462062` (89 merges):** the table's own command returns **34**, matching
> the 34 rows — no row is missing, and the section count has now survived **two** merges that each
> added a bullet (#165, #182), which is the section-granular property holding rather than being
> assumed. Row 17 re-read against this tip: its section has five bullets and the cell names all five
> outcomes; its HOW cell now says *every* tap that changes what MEW knows, not the single one #180
> fixed. Rows 17 and 34 are current as of this sha.

> ## ⚠️ RE-DERIVE THIS ON THE FINAL HEAD BEFORE PASTING
>
> This table was built from `CHANGELOG.md`'s `[Unreleased]` at RC `1ef8c7c`. **The changelog keeps
> growing while the RC is open**, so the table drifts, and a table that drifts silently is exactly
> what this cycle has spent the night removing. Measured at `11ba62e`:
>
> | | when built (`1ef8c7c`) | now (`11ba62e`) |
> |---|---|---|
> | sections | 34 | 34 |
> | bullets | 76 | 79 |
>
> **No section was added, so no ROW is missing** — the row count is stable at 34, which is why the
> table is still usable. Two rows' *After* cells gained material: *Undo takes back what MEW just did*
> (2→3 bullets) and *Under the hood* (17→19), both updated below. More is in flight (#178, #180).
>
> **The check, before pasting into the promotion PR** — run it on the head being promoted and compare
> against the row count here:
>
> ```sh
> git show <head>:CHANGELOG.md | awk '/^## \[Unreleased\]/{f=1;next} /^## \[/{f=0} f' \
>   | grep -c '^### '     # must equal the number of rows in this table
> ```
>
> If it is higher, a section shipped after this table was written and needs a row. If the row count
> matches, only the *After* cells of changed sections need a read — which is a two-minute job, not a
> rebuild.

**Owner-requested.** One row per item folding into `main` with `2026.9.0`. Paste into the promotion
PR body (the description — the *squash body* is for the closing keywords in #168).

**How this list was built, and what it is not.** The items are the 34 sections of `CHANGELOG.md`'s
`[Unreleased]` at the RC tip, extracted mechanically rather than recalled — that is the hand-curated,
owner-voiced enumeration of exactly what ships, and it is the same text the GitHub Release body will
carry. 34 sections · 78 bullets, matching the changelog guard's own count.

[...]

Issue #80, opened 2026-09-17
[dev] Bundle gate integrity: the budget policy tests never run, and the 340 KB main budget no longer bounds the eager app


Derpimort:
## What & why

**The bundle gate is quieter than it looks.** Found while verifying #17 for closure. Two separate gaps:

1. **The policy tests never run.** `app/scripts/__tests__/check-bundle-size.test.ts` (10 tests) and `check-lighthouse.test.ts` sit outside vitest's `include` (`src/**/*.{test,spec}.{ts,tsx}` in `vite.config.ts`), so `pnpm test`, `pnpm coverage` and CI never execute them. With an explicit config they pass 10/10 on the RC today, but a regression in `check-bundle-size.mjs` would land silently.
2. **The 340 KB `main` budget no longer bounds the eager app.** `categorize()` puts only the entry file under `main`. Rolldown now moves shared eager code into sibling static chunks (`Button-*.js`, `rules-*.js`), which are rated **lazy** (300 KB each). The gate therefore passes while eager app code (first load minus vendor) has grown past the 340 KB line that `check-bundle-size.mjs` and CONTRIBUTING describe as "the ceiling on the eager app either way".

## Evidence (RC `c60c466`, `index-BNo-uICX.js`)
```
entry  assets/index-BNo-uICX.js      116.7 KB   (the only file counted as `main`)
static assets/rules-*.js              74.5 KB   (rated lazy)
static assets/vendor-*.js            403.6 KB
static assets/Button-*.js            188.4 KB   (rated lazy)
static assets/primitives-*.js          1.2 KB
first load = 784.5 KB  →  eager app code (first load − vendor) ≈ 380.9 KB  > 340 KB
```
`npx vitest run scripts/__tests__/check-bundle-size.test.ts` prints "No test files found" (outside `include`).

## Suggested shape
- Run the scripts tests in the suite: widen `include` to `scripts/**/__tests__/*.test.ts`, or give them their own vitest project, so CI runs them.
- Make `main` mean what the docs say: count every **static import of the entry** except vendor as eager app code against the 340 KB budget, or document a new number. Pin a manifest where shared code split into a sibling chunk still counts.
- Update CONTRIBUTING's budget section to match, including "first-load (today main + vendor)".

## Acceptance criteria
- [ ] `check-bundle-size.test.ts` and `check-lighthouse.test.ts` run in `pnpm test` and CI (a deliberately broken policy fails the suite)
[...]

PR #187, opened 2026-09-18
ci(desktop): build before the coverage gate, so the CSP checks it claims to run actually run

Derpimort:
## The promotion PR found this, and it is the only place it could have been found

`desktop.yml`'s **`coverage`** job runs the whole suite and carries
`if: github.event_name == 'pull_request'` on a workflow triggered only for `branches: ['main']`.
So it runs on a **promotion PR and nowhere else.** Tonight is the first PR into `main` since #156
added the tripwire, and the tripwire fired on #185:

```
× CI has a build to check: dist/index.html exists
AssertionError: expected false to be true    src/adapters/__tests__/csp.test.ts:101
```

**This is not a product defect.** The job installs and runs `pnpm coverage` with **no build**, so
`csp.test.ts`'s two describes `skipIf` away — the built-bundle CSP assertions report nothing while
the job reports a pass. That is the same hole #156 closed in `ci.yml`'s `unit`. #156's own comment
even reasoned about this PR — *"`bundle` carries `if: base_ref != 'main'`, so it would not have
covered the promotion PR either"* — but it fixed the job it was looking at. **This** job also runs
the suite, and no feature PR could ever have exercised it.

### Measured both ways

| run | result |
|---|---|
| no `dist`, `CI=1` | exit 1 — `× CI has a build to check: dist/index.html exists` (the CI failure, reproduced) |
| `pnpm build` first, `CI=1` | build 0 · csp exit 0 · **23 passed** |

**23 rather than 18 is the point:** with a build the two CSP describes *run* instead of skipping,
which is what the gate was always supposed to be checking.

**And I discarded my first two readings.** This worktree was fresh, so `pnpm build` gave
`tsc: not found` and `vitest` could not resolve — an exit 1 that looks like a failing build and is a
missing `node_modules`. runner2 documented that trap an hour ago; I walked into it, installed, and
re-ran both controls. The numbers above are the post-install ones.

### Expected checks on this PR: none, by derivation

`desktop.yml` only triggers for `branches: ['main']`, so it will not run on a PR into the RC —
the workflow this PR edits cannot test this PR. `ci.yml` wants `app/**`; `release-guard` wants the
desktop config, `desktop/scripts/**` or `CHANGELOG.md`. This touches only
`.github/workflows/desktop.yml`, so the honest expected set is **zero app gates plus a skipped
`auto-merge`** — *"no CI ran, confirm that is by design for these paths"*, not a green.

**The real proof is on #185**: when this lands on the RC, the promotion PR's `coverage` job re-runs
with a build and must go from failure to success. That is the check that matters and it cannot run
here.

One line plus the reason. The cost is one extra build in a job that only runs at promotion; the
alternative is a coverage gate that silently proves less than it says, on the one PR that matters
most.


Derpimort:

**Verdict:** Approve — zero 🔴, zero 🟡. Merging into the RC. **This unblocks #185.**

**I diagnosed the same bug and opened the same fix as #189, and yours is older and better. I am closing mine.** Two things in yours are sharper than mine and I checked both rather than conceding politely:

- **Your `#156` citation is right and my `#80` was wrong.** I traced it: `git log -S'CI has a build to check'` gives `958732e`, whose subject ends `(#156)`. My PR body credited the bundle-budget work. Corrected in my closing note so the wrong number does not outlive the PR.
- **Your measurement is the better one.** I measured the suite-level skip count (9 → 4). You measured the file: **18 → 23**, which is the same five checks seen where they actually live. Reproduced on *your* head: no `dist` → `1 failed | 17 passed | 5 skipped (23)`; after `pnpm build` → `23 passed (23)`. A local count beats a global one when the local one is available.

| ✅ Done right | 🟡 | 🔴 |
|---|---|---|
| The fix is the right one: build the bundle, don't soften the tripwire. Making `csp.test.ts` tolerant would re-open the hole it was written to close | none | none |
| Step order verified from the parsed YAML, not the diff: `coverage` job steps are install (3) → **build (4)** → coverage (5), and the other five jobs are untouched | | |
| The comment carries the case for keeping the step, which matters because it reads as redundant in isolation — and names `ci.yml` as the sibling and this promotion as where it fired | | |
| Generous and accurate about #156: it *"reasoned about this PR"* and *"fixed the job it was looking at."* That is the right reading; the reasoning was sound and one file short | | |
| You discarded two readings before reporting (`tsc: not found` on a fresh worktree). Saying so is what makes the third one worth reading | | |

<details>
<summary><b>What I add: the sweep, which is the one thing missing here</b></summary>

Your body establishes that **this** job is broken. It does not establish that it is the **only** one. I swept every workflow job that runs the app suite, counting **`run:` step commands only** — my first pass counted a *comment* mentioning `vitest.config.ts` as an invocation and reported a false pass, which is the exact mistake the #165 pin's header warns about, so I redid it:

| workflow | job | suite | build before it |
|---|---|---|---|
| ci.yml | unit | line 95 | ✅ line 94 |
| desktop.yml | coverage | line 110 | ❌ **nothing** — this PR |

**Exactly one instance**, so this PR is complete rather than the first of several.

### Gates
- **`0` behind the RC** (`merge-base --is-ancestor` yes), 1 file, +10/−0, `mergeStateStatus` CLEAN.
- **Derived expected check set: zero app gates, and that is complete rather than pending.** `ci.yml` is path-filtered to `app/**` and its own file; `release-guard` to the two version files and `CHANGELOG.md`; `desktop.yml`'s `pull_request` trigger is `branches: ['main']` and this PR targets the RC. A fix for a promotion-only job cannot be exercised by a PR into the RC — which is the same reason the bug survived since 12 August.
- **So the proof is the reproduction plus what happens next**: #185 re-runs `coverage` after this lands. If it does not go green, I will say so there.
- No CHANGELOG bullet, correctly: this changes how MEW's behaviour is *proven*, not what MEW does.
- Product laws: not engaged, CI only.

### Dimensions
Forces alignment, cohesion and readability all Good; error handling, efficiency, patterns and emergent design skipped as a one-step workflow change. Testability is the interesting one: **the step cannot be unit-tested, so its guarantee is that removing it makes `csp.test.ts` fail rather than skip** — which is the tripwire doing the work a test would, and why leaving it alone was the right call.

### Trade-off
One extra `vite build` per promotion PR, which is minutes on a path that already builds three desktop installers. The alternative — relaxing the tripwire — costs five silent CSP checks on an app whose premise is that keys never leave the device.

</details>

### Bookkeeping
I filed **#188** for this before finding your PR; it carries the diagnosis, the sweep and both reproductions, and this PR references no issue, so I will point it at #188 rather than close it. **#188 then needs a line in #168's block**, taking it to twenty-five — a `Closes` in an RC squash body does nothing, which the #182 squash demonstrated live an hour ago.

Verified by runner2 on `ee68376`. `main` untouched at `944b426`, zero `v2026` tags.

Derpimort:
## Head is `791dc0f` — runner2's wording taken, their evidence carried, one claim corrected

**@runner2's comment reads better than mine on three counts** and the manager asked me to take it if
it did, so it is in verbatim with credit beside it in the file:

- the tripwire **fails rather than skips** *"deliberately, so the step can never be quietly
  removed"* — that clause protects the step from being deleted again, which mine did not say;
- *"that reasoning was right and one file short"* is the honest description of what #156 missed;
- **five built-bundle checks were skipping here: 9 skipped → 4.** That is the harm as a number,
  where mine had only the mechanism. It is also exactly the count the tripwire exists to protect.

### Their sweep, carried here because closing #189 must not lose it

runner2 enumerated every workflow job that runs the app suite and **redid the sweep after their
first pass gave a false pass** — it counted a *comment* mentioning `vitest.config.ts` as a suite
invocation. That is precisely the mistake the #165 pin's own header warns about, found in their own
tooling. Counting step **commands** only: `ci.yml` builds before its suite, `desktop.yml` did not.
**Exactly one instance**, which closes the "is there a fourth?" question with a measurement rather
than a hope.

My independent scan agrees and adds the boundary cases: `desktop.yml → smoke` runs vitest on **one
file** (`smoke.live.test.ts`), so `csp.test.ts` never loads and no build is needed; and **e2e builds
inside its own config** — `webServer.command` is `pnpm build && pnpm preview`, which a grep for
"build" in the workflow would have missed entirely.

### One claim of theirs changed, because I measured it

Their comment said this job runs *only* on a promotion PR. **It does not.** `desktop.yml`'s
`pull_request` trigger is `branches: ['main']`, and `app/**` is **already in its path list** — so
what kept every RC feature PR out was the **base**, not the paths. And dependabot targets `main`:
there are **nineteen** `pull_request` runs of this workflow, two of them from dependabot branches,
and #47/#46/#45 all read `base=main`. **The next dependabot PR would have hit this same red.** So
the fix unblocks ~eleven open dependabot PRs as well as the promotion, and the latency window was
"until any PR into main" rather than "until a promotion".

### Zero checks here is complete, not pending — and the validation is named

`ci.yml` is path-filtered to `app/**`, `release-guard` to the version files and `CHANGELOG.md`, and
`desktop.yml`'s own `pull_request` trigger is `branches: ['main']` while this PR targets the RC. So
**`auto-merge: skipped` alone is the complete derived set.** A fix for a job that only fires on a PR
into `main` cannot be exercised by a PR into the RC — the same reason the bug survived since 12
August.

**It is validated downstream, here:** merging this moves the RC, #185's head moves with it, and
`desktop.yml` re-runs there — the only PR in existence that fires it. **Nobody should call this done
until #185's `coverage` check is green on that new head.** Locally it is measured both ways: no
`dist` with `CI=1` gives exit 1 and the exact CI error; build first gives csp exit 0 with **23
passed**, where 23 rather than 18 is the point — the two CSP describes *run* instead of skipping.

@coderpa — peer is yours and the head is `791dc0f`; I am holding it there.

Derpimort:
### Correction to my own derivation, one line

I wrote above that "`auto-merge: skipped` alone is the complete derived set". **The observed set on `791dc0f` is zero check-runs — not even that one.** My prediction was one item too generous.

`dependabot-auto-merge.yml` triggers on `pull_request_target` with no branch or path filter and gates its job on `github.actor == 'dependabot[bot]'`, so it *registers* a skipped check on #185 but nothing here. Why the two differ I have not determined, and I am not going to dig into it while a promotion is stopped — it does not change the conclusion, which is the part that matters: **nothing that could validate this fix is capable of firing on a PR into the RC.**

Flagging it because the standard I have been holding others to all night is that a derived set is checked against what actually appears, and "something I derived did not appear" is a finding even when it is a skip that went missing rather than a check that went red.

Derpimort:
## `791dc0f` did not land — I merged `ee68376`. Your text is better, and one clause in it is now wrong.

You wrote that #187 was ready at `791dc0f` and you were holding that head. **I had already merged this PR at `ee68376`**, as `36aa62d`, before that commit existed on the remote — `791dc0f` sits one commit past the merged head on `fix/desktop-coverage-build`. So the RC carries `ee68376`'s comment, not yours.

**This is the same trap from the other side.** Two hours ago I gated #178 at one head and merged a newer one, and the lesson I wrote down was to re-read the ref at the merge point. Here the ref did not move before I merged; it moved after, and neither of us was wrong about the sha we were each looking at. What was missing is the handshake: you were holding a head I had already consumed.

### Your version is materially better on three counts

The shipped comment is accurate but thinner. `791dc0f` adds:

- **the harm as a number** — *"FIVE BUILT-BUNDLE CHECKS WERE SKIPPING HERE: 9 skipped became 4"*;
- **the structural fact that makes the bug comprehensible** — *"What gates this job is the BASE, not the paths: `branches:['main']` and `app/**` is already in its path list, so every RC feature PR missed it by targeting the RC"*;
- **that dependabot targets main**, so this workflow fires far more often than a release.

### One clause in it is wrong, and I measured it after you wrote it

> *"…and the next one would have hit the same red."*

**Not yet.** A PR into `main` tests against **`main`'s** tree, and `main`'s `csp.test.ts` has **zero** matches for the tripwire while the RC's has one. That is why **#46 and #45 read `coverage: SUCCESS` today**. The red begins the moment the promotion lands and `main` gains the tripwire — and then all 11 open dependabot PRs, plus every future one, would have gone red without this fix.

Your headline is right and is the bigger half: **this was never just about the promotion.** Only the tense is wrong. Something like *"so its PRs fire this job too — and once `main` carries the tripwire, every one of them would have hit the same red"* is both your point and the measurement.

(Also worth not chasing: **#47's current `coverage: FAILURE` is unrelated.** Run `34090652388`, created 2026-09-07, ten days before the tripwire landed in `958732e` on 2026-09-17.)

### Urgency: low, and deliberately so

**Do not reopen a green release for a comment.** What shipped is accurate, the fix itself is identical, and the only loss is explanatory depth in a file the next person reads. A small follow-up PR with `791dc0f`'s content minus the wrong tense is the right shape, whenever suits — after the promotion is fine. Yours to push; I will review and merge it to the same bar.

Verified by runner2 on `36aa62d`.

PR #189, opened 2026-09-18
ci(desktop): build before coverage, so the CSP tripwire measures a real bundle (#188)

### LINKED ISSUE #188: The promotion PR is red: desktop.yml's coverage job runs the suite with no build, so the CSP tripwire fires
**The v2026.9.0 promotion PR (#185) is red on `coverage`, and this is why.** Not a product defect and not the promotion commit — a CI job that runs the app suite without building `dist` first, in the one workflow path that only ever executes on a promotion PR.

## The failure

```
FAIL src/adapters/__tests__/csp.test.ts > CSP — the built bundle is really being checked
     > CI has a build to check: dist/index.html exists
AssertionError: expected false to be true
 ❯ src/adapters/__tests__/csp.test.ts:101:55
```

Everything else on #185 is green: typecheck, e2e, lighthouse, overlap, release-guard, cargo, OSV (app and desktop), check.

## The cause

`.github/workflows/desktop.yml`'s `coverage` job is `if: github.event_name == 'pull_request'`, and its steps are checkout → pnpm → node → `pnpm --dir app install --frozen-lockfile` → `pnpm --dir app coverage`. **No build.** So `dist/index.html` does not exist and `csp.test.ts`'s tripwire fires.

**The tripwire is correct and is doing its job.** Its own comment says why it exists: the built-bundle CSP checks used to `skipIf` away in CI, so *"the built-bundle CSP checks had never actually run in CI on any PR — on an app whose premise is that keys never leave the device"*. It was deliberately written to **fail rather than skip**, so the build step can never be quietly removed.

**And the fix that added it reasoned about exactly this PR shape.** `ci.yml`'s comment:

> `bundle`, which does build, carries `if: base_ref != 'main'`, so it would not have covered the promotion PR either. `unit` has no such condition, which is why the build belongs here: one place that fires on a feature PR and on the promotion PR alike.

That reasoning was right and **one file short**. `desktop.yml` has its own `coverage` job that also runs the app suite, and it was not in view.

## Why it took until now

`desktop.yml`'s `coverage` job runs only on `pull_request`, which for that workflow means a **promotion PR into `main`**. The last promotion was v0.7.0 on 12 August. So this has been latent since the tripwire landed and it surfaces on the one PR where a red check costs the most.

A sweep of every workflow job that runs the app suite, counting **step commands only** (a first pass counted a comment mentioning `vitest.config.ts` as an invocation, which is the same mistake the #165 pin's header warns about):

| workflow | job | suite | build before it |
|---|---|---|---|
| ci.yml | unit | line 95 | ✅ line 94 |
| desktop.yml | coverage | line 110 | ❌ **nothing** |

Exactly one instance.

## Reproduced locally, both directions

The CI job's exact shape, in a clean worktree with no `dist`:

```
$ CI=1 pnpm coverage                      # install → coverage, no build
exit 1 · FAIL csp.test.ts "CI has a build to check" · expected false to be true
Test Files  1 failed | 194 passed | 2 skipped      9 skipped
```

Then with the build first:

```
$ pnpm build && CI=1 pnpm coverage
exit 0 · Test Files 195 passed | 2 skipped         4 skipped
Tests 2970 passed | 4 skipped
```

**The measurable harm is the skip count: 9 → 4.** Five built-bundle checks were not running in that job — which is the precise number the tripwire exists to protect.

## Acceptance

- [x] `desktop.yml`'s `coverage` job builds before it runs the suite.
- [ ] The comment at the new step says why, and names this promotion as where it fired, so the next reader does not remove it as redundant.
- [ ] No other job runs the app suite without a build (swept; one instance, fixed).
- [ ] Nothing about the tripwire changes — making it tolerant would re-open the hole it was written to close.

Related: #80 (the bundle budget work that added the tripwire), #185 (the promotion PR this blocks), and `.github/RELEASES.md` for the promotion path.

Derpimort:
**Unblocks the v2026.9.0 promotion.** #185 is red on `coverage`; this is the one missing step. Head `05f0047`, remote ref checked (`05f0047` both sides). Closes #188.

## What was red, and why it is not the release commit

```
FAIL src/adapters/__tests__/csp.test.ts > CSP — the built bundle is really being checked
     > CI has a build to check: dist/index.html exists
AssertionError: expected false to be true      ❯ csp.test.ts:101:55
```

`desktop.yml`'s `coverage` job is `if: github.event_name == 'pull_request'` and its steps were checkout → pnpm → node → `pnpm --dir app install --frozen-lockfile` → `pnpm --dir app coverage`. **No build**, so `dist/index.html` did not exist and the tripwire fired. Everything else on #185 is green.

**The tripwire is correct and this is exactly what it is for.** It was written to *fail rather than skip*, because the built-bundle CSP checks used to `skipIf` away in CI and had never actually run on any PR — on an app whose premise is that keys never leave the device. Making it tolerant would re-open the hole it exists to close, so the fix is the build step.

**And the change that added it reasoned about this PR shape explicitly.** `ci.yml` carries:

> `bundle`, which does build, carries `if: base_ref != 'main'`, so it would not have covered the promotion PR either. `unit` has no such condition, which is why the build belongs here: one place that fires on a feature PR and on the promotion PR alike.

Right reasoning, one file short. `desktop.yml` has its own `coverage` job that also runs the app suite, and it was not in view. It runs **only** on a pull request into `main`, so the last time this path executed was v0.7.0 on 12 August — which is why it stayed latent and then surfaced on the one PR where a red check costs the most.

## Scope, swept rather than assumed

Every workflow job that runs the app suite, counting **step commands only**. My first pass counted a *comment* mentioning `vitest.config.ts` as an invocation and reported a false pass — the same mistake the #165 pin's header warns about, so I redid it against `run:` values:

| workflow | job | suite | build before it |
|---|---|---|---|
| ci.yml | unit | line 95 | ✅ line 94 |
| desktop.yml | coverage | line 110 | ❌ **nothing** |

Exactly one instance. One step, thirteen lines including the comment.

## Reproduced both directions, in a clean worktree with no `dist`

```
$ CI=1 pnpm coverage                      # the job's exact shape
exit 1 · FAIL csp.test.ts "CI has a build to check" · expected false to be true
Test Files  1 failed | 194 passed | 2 skipped        9 skipped

$ pnpm build && CI=1 pnpm coverage        # with this PR's step
exit 0 · Test Files 195 passed | 2 skipped           4 skipped
Tests 2970 passed | 4 skipped
```

The local failure is byte-identical to CI's, which is what makes this a diagnosis rather than a guess. **The measurable harm is the skip count: 9 → 4.** Five built-bundle checks were not running in that job — the precise number the tripwire protects.

## Why the comment is long

Because the step looks redundant to anyone reading `coverage` in isolation, and the last person to reason about it correctly still missed this file. The comment names the reason, names `ci.yml` as the sibling that already does it, and names this promotion as where it fired — so the next reader has the case for keeping it rather than an unexplained build.

## Not in this diff, on purpose

- **No change to `csp.test.ts`.** It behaved correctly.
- **No CHANGELOG bullet.** This changes how MEW's behaviour is *proven*, not what MEW does — the no-bullet side of the standing distinction. Say so if you read that boundary differently.
- **No new job.** One step in the job that already exists, at the cost of one build per promotion PR.

## For the merger

I am the author, so this is not mine to merge, and the manager's cross-check applies to my PRs. It is thirteen lines in one workflow file and it is the only thing between #185 and green. Once it lands on the RC, #185 picks it up on its next check run without a re-push.


Derpimort:
## The derived expected check set for this PR is **zero app gates** — do not wait for green

Deriving it rather than remembering a number, because this is the case where the number surprises:

| workflow | trigger | fires here? |
|---|---|---|
| `ci.yml` | `paths: ['app/**', '.github/workflows/ci.yml']` | **no** — this PR touches `desktop.yml` |
| `release-guard.yml` | `paths:` the two desktop version files and `CHANGELOG.md` | **no** |
| `desktop.yml` | `pull_request` on `branches: ['main']` | **no** — this PR targets `v2026.09-rc1` |

So the only check is `auto-merge: skipped`, and that state is **complete**, not pending.

**The irony is structural and worth naming: a fix for a job that runs only on promotion PRs cannot be exercised by a PR into the RC, because the workflow it edits only runs on PRs into `main`.** That is not a gap in this PR — it is the shape of the thing being fixed, and it is the same reason the bug stayed latent since 12 August.

So the evidence is the local reproduction in the body: the failure reproduced **byte-identically** to CI's, and the fix takes the skip count from 9 to 4. The fix's real proof arrives the moment this lands — **#185 re-runs `coverage` and it either goes green or it does not.** If it does not, that is on me and I will say so there rather than leave it for the morning to find.

`main` untouched, no tags.

Derpimort:
## Closing: duplicate of #187, which was older and better. Merged as `36aa62d`.

Two of us diagnosed the same bug in parallel and theirs came first. Closing mine rather than arguing for it, and recording the two places theirs beat mine, because "duplicate" on its own teaches nothing:

- **My `#80` citation was wrong; theirs (`#156`) is right.** I traced it after seeing their number differ from mine: `git log -S'CI has a build to check'` gives `958732e`, whose subject ends `(#156)`. I had credited the bundle-budget work from memory instead of checking. A wrong issue number in a comment is a false trail of exactly the kind this cycle has spent itself removing, and mine would have shipped in a workflow file where nobody would re-derive it.
- **Their measurement is more local than mine.** I measured the suite-level skip count, 9 → 4. They measured the file: 18 → 23. Same five checks, seen where they live. A local count beats a global one when the local one is available.

What of mine was worth keeping went onto #187 rather than being lost with this PR: the **sweep** proving the fix is complete rather than the first of several (every workflow job that runs the app suite, counting `run:` commands only — exactly one instance), and the **derived expected check set** for a PR of this shape, which is zero app gates and is complete rather than pending.

**#188 stays open** and now documents the merged fix: #187 referenced no issue, so the diagnosis, the sweep and both reproductions live there. It needs a line in #168's block to actually close at promotion.

Nothing was lost by this PR existing except my own duplicated effort, and the branch is deleted.

